PR #201, opened 2014-12-23
Fix issue when an operator follows equals()

### LINKED ISSUE #181: Expr equals() conflicts with other operators
Originally reported in: doctrine/mongodb-odm#917

User attempted to use `->equals('foo')` and `->notEquals('bar')` on the same field, which clashes. If `equals()` is used first, other operators will throw a PHP warning because the field cannot be access as an array (or probably worse undefined behavior if it actually is an array). Inversely, `equals()` would simply overwrite previously applied operations.

alcaeus:
This fixes #181 by overwriting the existing query with an empty array to avoid invalid string offsets or invalid array accesses when calling any query operator after having called ```equals()``` on the field.

### DIFF
--- a/lib/Doctrine/MongoDB/Query/Expr.php
+++ b/lib/Doctrine/MongoDB/Query/Expr.php
@@ -713,6 +713,10 @@ class Expr
     public function operator($operator, $value)
     {
         if ($this->currentField) {
+            if (isset($this->query[$this->currentField]) && !is_array($this->query[$this->currentField])) {
+                $this->query[$this->currentField] = array();
+            }
+
             $this->query[$this->currentField][$operator] = $value;
         } else {
             $this->query[$operator] = $value;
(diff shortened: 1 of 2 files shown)


jmikola:
@alcaeus: I'm unsure if overriding the previous equality criteria is the best course of action here. Ideally, we'd wrap the previous equality with `$eq`, but that isn't supported until [server 2.8](http://docs.mongodb.org/v2.8/reference/operator/query/eq/).

For a backwards compatible approach, we could wrap the original equality criteria in a single-element `$in` array; however, your particular test is evidence that overwriting is still possible. This is also more likely to elicit WTFs from users.

Would it be more straightforward to use `$eq` (the proper solution) and invite an exception if the server doesn't support it?

alcaeus:
To be honest, I haven't given that much thought, but I'd go with the backwards compatible approach.
It gets me wondering: are there use cases where there would be multiple operations in a single field so that overwriting is a bad idea?

jmikola:
> are there use cases where there would be multiple operations in a single field so that overwriting is a bad idea?

Offhand, I can't think of any. I'll make the change and merge it with that behavior then. Thanks!

jmikola:
Manually merged in https://github.com/doctrine/mongodb/commit/724be1d4e864cfbe098be6beb3d610bf492c1516. I included your original commit but added logic to attempt merging existing equality criteria in https://github.com/doctrine/mongodb/commit/5dfc371713a939068ee99c50f40148fadff215ae. I also added some tests to document the edge cases.

Thanks!

alcaeus:
Awesome, thank you!

PR #207, opened 2015-02-24
Bump doctrine/common dependency

jmikola:


.gitignore
```diff
@@ -1,5 +1,2 @@
 phpunit.xml
-vendor
-composer.lock
```
stof: why commiting the lock file in the library ?

jmikola: I talked with @naderman about this at FOSDEM. It's ignored for projects that depend on ODM, so it doesn't affect our users, but it gives developers consistency for reproducing test results (also the "build status" icon).

BernardoSilva:
Commiting the lock file will force dependencies to be locked for that specific version.
If you're locking the dependencies to a specific revision it can cause issues if we depend directly on `doctrine/mongodb`.

naderman:
Composer does not consider the content of a dependency's lock file, only the lock file in the root project itself is ever used.

PR #209, opened 2015-03-30
Add a common cursor interface

### LINKED ISSUE #208: EagerCursor does not support sorting
The ```Doctrine\MongoDB\EagerCursor``` class wraps ```Doctrine\MongoDB\Cursor```, but does not offer methods like ```sort()```. Instead, they must be called using ```$eagerCursor->getCursor()->sort()```. While this works, it also means that the two cursor classes are not interchangeable.
Note: this trickles down to the ODM as well, since the ODM\EagerCursor class also does not support sorting which in turn causes other errors. IMO, the EagerCursor class should also offer the sort methods.

alcaeus:
This PR aims to fix #208 by adding a common ```CursorInterface``` which contains most methods from ```MongoCursor``` and aims to have ```Cursor``` and ```EagerCursor``` provide the same functionality. Note that ```EagerCursor``` will delegate all calls to the wrapped ```Cursor``` instance. It's up to the user to ensure the underlying cursor object has not been opened by calling ```current``` or ```next```.

### DIFF
--- a/lib/Doctrine/MongoDB/CursorInterface.php
+++ b/lib/Doctrine/MongoDB/CursorInterface.php
@@ -53,7 +53,6 @@ interface CursorInterface extends Iterator
      *
      * @see http://php.net/manual/en/countable.count.php
      * @see http://php.net/manual/en/mongocursor.count.php
-     * @param boolean $foundOnly
      * @return integer
      */
     public function count();
--- a/lib/Doctrine/MongoDB/CursorInterface.php
+++ b/lib/Doctrine/MongoDB/CursorInterface.php
@@ -1,0 +1,198 @@
+<?php
+/*
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS
+ * "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT
+ * LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR
+ * A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+ * OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+ * SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+ * LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+ * DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+ * THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+ * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\MongoDB;
+
+use Doctrine\MongoDB\Util\ReadPreference;
+
+/**
+ * Wrapper for the PHP MongoCursor class.
+ *
+ * @since  1.0
+ * @author alcaeus <alcaeus@alcaeus.org>
+ */
+interface CursorInterface extends Iterator
+{
+    /**
+     * Wrapper method for MongoCursor::addOption().
+     *
+     * @see http://php.net/manual/en/mongocursor.addoption.php
+     * @param string $key
+     * @param mixed $value
+     * @return self
+     */
+    public function addOption($key, $value);
+
+    /**
+     * Wrapper method for MongoCursor::batchSize().
+     *
+     * @see http://php.net/manual/en/mongocursor.batchsize.php
+     * @param integer $num
+     * @return self
+     */
+    public function batchSize($num);
+
+    /**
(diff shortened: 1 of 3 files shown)


lavoiesl:
You need tu have an empty line before `return` statements.

To avoid duplication, I would replace all docblocks of methods defined in the interface by `{@inheritdoc}`

lib/Doctrine/MongoDB/CursorInterface.php
```diff
@@ -0,0 +1,198 @@
...
+
+    /**
+     * Wrapper method for MongoCursor::count().
+     *
+     * @see http://php.net/manual/en/countable.count.php
+     * @see http://php.net/manual/en/mongocursor.count.php
+     * @param boolean $foundOnly
+     * @return integer
+     */
+    public function count();
```
jmikola: We have `@param boolean $foundOnly` above, which should default to `false`. Shouldn't it also appear in the method prototype?

lib/Doctrine/MongoDB/EagerCursor.php
```diff
@@ -65,6 +65,7 @@ public function __construct(Cursor $cursor)
     public function count()
     {
         $this->initialize();
+
         return count($this->data);
```
jmikola: If we change the `count()` prototype to `count($foundOnly = false)`, as suggested above, we'll need to add the same here. Unfortunately, a false value for that option doesn't really apply to EagerCursor, since it inherently only counts the found documents after initialization. I'm not sure what the best solution is for that beyond documentation.

alcaeus: I dislike adding a feature (as minor as it might be) in an interface and then violate this interface in a child class. In this case, I'd rather remove the ```$foundOnly``` parameter in the cursor interface and only add it in those cursors that actually support it.

jmikola: :+1:

alcaeus:
@jmikola I gave ```EagerCursor::count()``` some more thought after running into a couple of issues today. Currently, ```EagerCursor::count()``` and ```Cursor::count()``` behave very different, even though they are supposed to just wrap ```MongoCursor```. I'll let the code speak volumes here:
```
    public function testFunctionalCursorCount()
    {
        $db = $this->conn->selectDatabase(self::$dbName);

        /** @var Collection $coll */
        $coll = $db->selectCollection('users');

        $document = array('user' => 'jonwage');
        $coll->insert($document);

        $document = array('user' => 'jmikola');
        $coll->insert($document);

        $document = array('user' => 'malarzm');
        $coll->insert($document);

        $document = array('user' => 'alcaeus');
        $coll->insert($document);

        $builder = $coll->createQueryBuilder();
        $builder->skip(1)->limit(4);
        $cursor = $builder->getQuery()->getIterator();

        $this->assertInstanceOf('Doctrine\MongoDB\Cursor', $cursor);
        $this->assertEquals(4, $cursor->count());
        $this->assertEquals(3, $cursor->count(true));

        $eagerCursor = $builder->eagerCursor(true)->getQuery()->getIterator();

        $this->assertInstanceOf('Doctrine\MongoDB\EagerCursor', $eagerCursor);
        $this->assertEquals(4, $eagerCursor->count());
        $this->assertEquals(3, $eagerCursor->count(true));
    }
```

This test will fail because ```$eagerCursor->count()``` will return 3 instead of 4.

Suggestion: Use ```$foundOnly = false``` for ```EagerCursor``` as well. This way we could implement the same behavior as ```Cursor::count()``` with minimal overhead.

jmikola:
> Use `$foundOnly = false` for EagerCursor as well.

Wouldn't that require that EagerCursor internally fetch and store results with skip/limit ignored? That seems problematic if the user is purposefully using limit (I suppose skip isn't too relevant here) to conserve memory usage or limit the returned server response.

I would think `$foundOnly` would have to always be true (not just default to it); however, this does create an unfortunate API difference with the regular Cursor.

alcaeus:
> Wouldn't that require that EagerCursor internally fetch and store results with skip/limit ignored?
Actually, no. It would still fetch and store results with skip/limit as usual. ```count()``` however would behave different if ```$foundOnly``` is false. In that case, instead of counting array elements (and potentially initializing the result array) it would fetch and store the total result count from the cursor. That way you'd have one query to fetch results and at most one extra count command if you were to call ```count()``` with the default parameter.

jmikola:
> That way you'd have one query to fetch results and at most one extra count command if you were to call `count()` with the default parameter.

I'd consult with @jwage first, but I think this defeats the original intention of EagerCursor in that should would avoid any database interaction after initialization. There's the whole separate issue of `MongoCursor::count()` generally being a bad bit of API design, since it invokes a command and is prone to returning an "incorrect" count (if we consider `iterator_count()` to be authoritative).

Per your original comment earlier in the PR diff, when I originally suggested adding `$foundOnly` to the interface prototype:

> I dislike adding a feature (as minor as it might be) in an interface and then violate this interface in a child class. In this case, I'd rather remove the $foundOnly parameter in the cursor interface and only add it in those cursors that actually support it.

In this case, can't we say that EagerCursor doesn't support a `$foundOnly` argument and only counts the documents it knows about? This makes its behavior equivalent to `$foundOnly = true`, which just happens not to be MongoCursor's and our Cursor class' default behavior.

alcaeus:
You're right. Maybe we can simplify this at a later date, at this point it would also be a BC break and thus a no-go for 1.2. We'll leave as is.

jmikola:
LGTM.

lib/Doctrine/MongoDB/CursorInterface.php
```diff
@@ -0,0 +1,197 @@
...
+ */
+
+namespace Doctrine\MongoDB;
+
+use Doctrine\MongoDB\Util\ReadPreference;
+
+/**
+ * Wrapper for the PHP MongoCursor class.
+ *
+ * @since  1.0
```
jmikola: @alcaeus: I believe this should be corrected to 1.2, yes?

alcaeus: I did so in #225 :)

jmikola: :+1:

PR #212, opened 2015-05-01
Query builder support for new update operators in MongoDB 2.6

### LINKED ISSUE #136: Query builder support for new update operators in MongoDB 2.6
See: http://docs.mongodb.org/master/release-notes/2.6/#new-update-operators

alcaeus:
This PR adds support for the ```$bit```, ```$currentDate```, ```$max```, ```$min```, ```$mul``` and ```$position``` update operators added in MongoDB 2.6.

Fixes #136.

### DIFF
--- a/lib/Doctrine/MongoDB/Query/Expr.php
+++ b/lib/Doctrine/MongoDB/Query/Expr.php
@@ -591,6 +591,21 @@ class Expr
     }
 
     /**
+     * Updates the value of the field to a specified value if the specified value is greater than the current value of the field.
+     *
+     * @see Builder::max()
+     * @see http://docs.mongodb.org/manual/reference/operator/update/max/
+     * @param mixed $value
+     * @return self
+     */
+    public function max($value)
+    {
+        $this->requiresCurrentField();
+        $this->newObj['$max'][$this->currentField] = $value;
+        return $this;
+    }
+
+    /**
      * Set the $maxDistance option for $near or $nearSphere criteria.
      *
      * This method must be called after near() or nearSphere(), since placement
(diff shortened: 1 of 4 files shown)


malarzm:
LGTM :)

lib/Doctrine/MongoDB/Query/Expr.php
```diff
@@ -162,6 +162,75 @@ public function all(array $values)
...
+
+    /**
+     * Sets the value of the current field to the current date, either as a date or a timestamp.
+     *
+     * @see Builder::currentDate()
+     * @see http://docs.mongodb.org/manual/reference/operator/update/currentDate/
+     * @param bool $useTimestamp
+     * @return self
+     */
+    public function currentDate($useTimestamp = false)
```
jmikola: @alcaeus: Sorry for the late comment, but what do you think about changing this to `$type = 'date'`? We should generally discourage users from using the BSON timestamp type, since it's actually more than a UNIX timestamp. A boolean option makes it very easy for them to inadvertently use it. This also keeps the method flexible if the `$currentDate` operator changes down the line to support additional types (e.g. date with a timezone).

alcaeus: I wasn't sure whether it would be needed since I've never used a timestamp. Since we offer a timestamp type in ODM I'd still allow the use of ```$currentDate``` with ```$type = 'timestamp'```. Since the next logical step is adding those operators to the ODM builder we could document this the same way we do for the ```@Timestamp``` type: https://doctrine-mongodb-odm.readthedocs.org/en/latest/reference/annotations-reference.html#timestamp

I admit that the bool parameter is not very future proof - maybe use a ```$type``` parameter with constants for timestamp and date that can be extended should the operator support more types?

jmikola: The `@Timestamp` mapping is basically stuck in ODM now (and I suppose it has some relevance since MongoTimestamp is a legitimate class the driver could return).

There is a similar need if we were to ever introduce constants for `Expr::type()` (see: [here](https://github.com/doctrine/mongodb/blob/c3c029cab1917afb8fa7cc41ef3dc328e8a852fb/lib/Doctrine/MongoDB/Query/Expr.php#L1126), where we currently use hard-coded strings), so any decision to introduce constants should be mindful to use a prefix so not to cause a conflict down the line (e.g. using `CURRENTDATE_DATE` instead of just `DATE`). I don't think we need to introduce constants for this, though. Although `"date"` is a "magic string", it's something 99% of users would be using, so I think we can get by with having it as a default value.

If you're sold on having constants for this, though, I won't object. Just prefix them :)

PR #213, opened 2015-05-02
Add aggregation builder

alcaeus:
```Doctrine\MongoDB\Collection``` has an ```aggregate()``` method that will run an aggregation pipeline against the collection. Unfortunately, the pipeline itself has to be defined with usual arrays:
```
$expectedPipeline = array(
    array(
        '$geoNear' => array(
            'near' => $point,
            'spherical' => true,
            'distanceField' => 'distance',
            'query' => array(
                'hasCoordinates' => array('$exists' => true),
                'username' => 'foo',
            ),
            'num' => 10
        )
    ),
    array('$match' =>
        array(
            'group' => array('$in' => array('a', 'b'))
        )
    ),
    array('$unwind' => 'a'),
    array('$unwind' => 'b'),
    array('$redact' =>
        array(
            '$cond' => array(
                'if' => array('$lte' => array('$accessLevel', 3)),
                'then' => '$$KEEP',
                'else' => '$$REDACT',
            )
        )
    ),
    array('$project' =>
        array(
            '_id' => false,
            'user' => true,
            'amount' => true,
            'invoiceAddress' => true,
            'deliveryAddress' => array(
                '$cond' => array(
                    'if' => array(
                        '$and' => array(
                            array('$eq' => array('$useAlternateDeliveryAddress', true)),
                            array('$ne' => array('$deliveryAddress', null))
                        )
                    ),
                    'then' => '$deliveryAddress',
                    'else' => '$invoiceAddress'
                )
            )
        )
    ),
    array('$group' =>
        array(
            '_id' => '$user',
            'numOrders' => array('$sum' => 1),
            'totalAmount' => array('$sum' => '$amount'),
            'avgAmount' => array('$avg' => '$amount')
        )
    ),
    array('$sort' => array('totalAmount' => 0, 'numOrders' => -1, 'avgAmount' => 1)),
    array('$limit' => 5),
    array('$skip' => 2),
    array('$out' => 'collectionName')
);
```

@jmikola has previously said (https://github.com/doctrine/mongodb/issues/89#issuecomment-14576474):
> The real feature I had in mind was something integrated into the query builder, but I suppose that can wait.

That was some time ago, and since I recently needed to write some big aggregations I decided to give the whole thing a shot. I've decided against integrating it in the query builder and instead made a dedicated aggregation builder. This helps keeping both functionalities apart.

The idea behind this is best explained with a sample pipeline:
```
$builder = $collection->createAggregationBuilder();
$builder
    ->geoNear($point)
        ->distanceField('distance')
        ->limit(10) // Limit is applied on $geoNear
        ->field('hasCoordinates')
        ->exists(true)
        ->field('username')
        ->equals('foo')
    ->match()
        ->field('group')
        ->in(array('a', 'b'))
    ->unwind('a')
    ->unwind('b')
    ->redact()
        ->cond(
            $builder->operator()->lte('$accessLevel', 3),
            '$$KEEP',
            '$$REDACT'
        )
    ->project()
        ->excludeIdField()
        ->includeFields(array('user', 'amount', 'invoiceAddress'))
        ->field('deliveryAddress')
        ->cond(
            $builder->operator()
                ->addAnd($builder->operator()->eq('$useAlternateDeliveryAddress', true))
                ->addAnd($builder->operator()->ne('$deliveryAddress', null)),
            '$deliveryAddress',
            '$invoiceAddress'
        )
    ->group()
        ->field('_id')
        ->value('$user')
        ->field('numOrders')
        ->sum(1)
        ->field('totalAmount')
        ->sum('$amount')
        ->field('avgAmount')
        ->avg('$amount')
    ->sort('totalAmount')
    ->sort(array('numOrders' => 'desc', 'avgAmount' => 'asc')) // Multiple subsequent sorts are combined into a single stage
    ->limit(5)
    ->skip(2)
    ->out('collectionName');
```

TODOs:
- [x] ```Aggregation\Expr``` object
- [x] redact stage
- [x] group stage
- [x] project stage
- [x] nested documents in project and group stages with appropriate methods
- [x] Tests
  - [x] Tests for Aggregation\Expr
  - [x] Tests for Aggregation\Builder
  - [x] Tests for individual stages
- [ ] Fix Scrutinizer issues

### CODE REVIEW

**Comment by gruberro** [comment]
:+1:

File: lib/Doctrine/MongoDB/Aggregation/Expr.php [hunk 1]
```diff
@@ -0,0 +1,890 @@
...
+    /**
+     * Adds numbers together or adds numbers and a date. If one of the arguments is a date, $add treats the other arguments as milliseconds to add to the date.
+     *
+     * The arguments can be any valid expression as long as they resolve to either all numbers or to numbers and a date.
+     *
+     * @see http://docs.mongodb.org/manual/reference/operator/aggregation/add/
+     * @param mixed|self $expression1
+     * @param mixed|self $expression2
+     * @param mixed|self $expression3,... Additional expressions
+     * @return $this
```
malarzm: Cosmetics: we seem to use `@return self` in QueryBuilder

bicatu:
Hi any idea when this will be available?  Do you know if it will support DBRef as part of the query?

alcaeus:
@bicatu No idea when it will be ready. For now it supports all things I could think of when it comes to aggregation, I need to add unit tests which I won't be able to do until in a few weeks. Until then, if you're brave you can play around and see what complicated or larger aggregations feel like.

As for DBRef, it supports it as rudimentary as MongoDB allows within aggregation:
* You can add a query for ```field.$id``` in the ```$match``` stage
* You can use the entire field in a ```$project```, ```$group``` or ```$redact``` stage like so ```$group: { _id: { '$fieldWithReference' } }```
* You can **not** specifically use any nested fields of a DBRef object due to the $ being reserved as operator, so this won't work: ```$group: { _id: { '$fieldWithReference.$id' } }```

The last one is a limitation imposed by MongoDB that we can't circumvent.

bicatu:
Hi, I am trying your fork/branch and I can use for all but my DBRef queries.

Ex. The following match works in the console

db.collection.aggregate([ {$match: {member: DBRef("Member", ObjectId("5566706ea10fdb0ddc0041a7"), "doctrine_odm")}}])

but when I try from the code I get zero results.

```php
$builder = $collection->createAggregationBuilder();
        $builder->match()
            ->field('member')
            ->equals(\MongoDBRef::create("Member", "5566706ea10fdb0ddc0041a7", "doctrine_odm"));
```

solocommand:
@bicatu Shouldn't the second param to MongoDBRef::create be wrapped in an ObjectId constructor?

bicatu:
Indeed, after I added the new \MongoId() it worked.  Thanks.

So now really we just have to wait for it to be merged :)

:+1:

bicatu:
Hi, beside time is there anything preventing this from being merged?

jmikola:
@bicatu: I spoke with @alcaeus earlier this week about reviewing this and several other outstanding PRs so we can triage for a 1.2.0 release. I hope to get to this and the Cursor interface stuff before the weekend.

alcaeus:
Well, besides @jmikola and his all-seeing eye the only thing missing from my end are tests. I was on vacation the past two weeks which is why nothing has moved along. I'll pick this up again in the next couple of days™.

bicatu:
@alcaeus let me know if I can help.

alcaeus:
@bicatu You can test it with the aggregation queries you are currently running against your database. I'm especially interested on feedback on how it feels to use the aggregation builder. The goal was to make the builder work like the query builder as much as possible, with the exception that aggregation queries are by nature more complex than regular lookups.
Other than that I'll just push through with unit test - of course you're welcome to help with those if you'd like ;)

lib/Doctrine/MongoDB/Aggregation/Expr.php
alcaeus: I changed it to ```@return self``` now. For the record, in many IDEs this makes a difference in auto-completion. Using ```$this``` instead of ```self``` provides correct auto-completion when extending other classes.

lib/Doctrine/MongoDB/Aggregation/Builder.php
```diff
@@ -0,0 +1,262 @@
...
+     */
+    protected $collection;
+
+    /**
+     * @var Stage[]
+     */
+    protected $stages;
+
+    /**
+     * Create a new query builder.
```
malarzm: query = aggregation?

lib/Doctrine/MongoDB/Aggregation/Builder.php
```diff
@@ -0,0 +1,262 @@
...
+
+    /**
+     * Returns a certain stage from the pipeline
+     *
+     * @param integer $index
+     * @return Stage
+     */
+    public function getStage($index)
+    {
+        if (!isset($this->stages[$index])) {
```
malarzm: CS: can you add spaces around `!` (here and everywhere else, won't be commenting on them ;) )?

lib/Doctrine/MongoDB/Aggregation/Expr.php
```diff
@@ -0,0 +1,890 @@
...
+
+    /**
+     * Ensure that a current field has been set.
+     *
+     * @throws LogicException if a current field has not been set
+     */
+    protected function requiresCurrentField()
+    {
+        if (!$this->currentField) {
+            throw new LogicException('This method requires you set a current field using field().');
```
malarzm: Would it be better to use something like "Expression must be associated with field, please use field() first." or something similar?

lib/Doctrine/MongoDB/Aggregation/Stage/Match.php
```diff
@@ -0,0 +1,604 @@
...
+    {
+        return array(
+            '$match' => $this->query->getQuery()
+        );
+    }
+
+    /**
+     * Add an $and clause to the current query.
+     *
+     * You can create a new expression using the {@link Builder::expr()} method.
```
malarzm: Should this (and upcoming `{@link}` point to `Builder` or to this class?

malarzm:
Ufff that was huge, definitely :+1: for doing this! LGTM beside few small things I've pointed out :)

lib/Doctrine/MongoDB/Aggregation/Builder.php
```diff
@@ -0,0 +1,262 @@
...
+     * @param array $options
+     * @return Iterator
+     */
+    public function execute($options = array())
+    {
+        return $this->collection->aggregate($this->getPipeline(), $options);
+    }
+
+    /**
+     * Outputs documents in order of nearest to farthest from a specified point. You can only use this as the first stage of a pipeline.
```
jmikola: Line wrapping?

This doesn't have to fit on one line, but it should probably be a single sentence. You can add the note about this only being allowed as the first stage in its own paragraph (either between this and the next, or as the third).

lib/Doctrine/MongoDB/Aggregation/Builder.php
```diff
@@ -0,0 +1,262 @@
...
+    /**
+     * Returns a certain stage from the pipeline
+     *
+     * @param integer $index
+     * @return Stage
+     */
+    public function getStage($index)
+    {
+        if (!isset($this->stages[$index])) {
+            throw new \InvalidArgumentException("Could not find stage with index {$index}.");
```
jmikola: I think OutOfRangeException is more appropriate here. It's a LogicException, like InvalidArgumentException and unlike OutOfBoundsException, which is a RuntimeException.

lib/Doctrine/MongoDB/Aggregation/Builder.php
```diff
@@ -0,0 +1,262 @@
...
+     */
+    public function getPipeline()
+    {
+        $pipeline = array();
+
+        foreach ($this->stages as $stage) {
+            $pipeline[] = $stage->getExpression();
+        }
+
+        return $pipeline;
```
jmikola: Ripe opportunity to make @Ocramius happy with some functional programming.

```
return array_map(
    function(Stage $stage) { return $stage->getExpression(); },
    $this->stages
);
```

lib/Doctrine/MongoDB/Aggregation/Builder.php
```diff
@@ -0,0 +1,262 @@
...
+     *
+     * @see http://docs.mongodb.org/manual/reference/operator/aggregation/sort/
+     *
+     * @param array|string $fieldName Field name or array of field/order pairs
+     * @param integer|string $order   Field order (if one field is specified)
+     * @return Stage\Sort
+     */
+    public function sort($fieldName, $order = null)
+    {
+        return $this->addStage(new Stage\Sort($this, $fieldName, $order));
```
jmikola: Your OP includes a comment that "Multiple subsequent sorts are combined into a single stage". Is that implemented elsewhere, or was it just an original idea. MongoDB is already going to optimize the pipeline, so I don't think we need to attempt any optimizations.

lib/Doctrine/MongoDB/Aggregation/Builder.php
```diff
@@ -0,0 +1,262 @@
...
+
+        foreach ($this->stages as $stage) {
+            $pipeline[] = $stage->getExpression();
+        }
+
+        return $pipeline;
+    }
+
+    /**
+     * Groups documents by some specified expression and outputs to the next stage a document for each distinct grouping.
```
jmikola: Similar comment as with `geoNear()`. Even if one sentence, we can wrap this.

I'll refain from leaving redundant comments on the other docblocks below.

lib/Doctrine/MongoDB/Aggregation/Expr.php
```diff
@@ -0,0 +1,890 @@
...
+     */
+    public function divide($expression1, $expression2)
+    {
+        return $this->operator('$divide', array($expression1, $expression2));
+    }
+
+    /**
+     * Compares two values and returns:
+     * true when the values are equivalent.
+     * false when the values are not equivalent.
```
jmikola: I realize this is quoting the docs, but I think we can rewrite to a single line for docblock usage. Perhaps: "Compares two expressions and returns whether they are equivalent."

lib/Doctrine/MongoDB/Aggregation/Expr.php
```diff
@@ -0,0 +1,890 @@
...
+     * @param mixed|self $expression2
+     * @return self
+     */
+    public function eq($expression1, $expression2)
+    {
+        return $this->operator('$eq', array($expression1, $expression2));
+    }
+
+    /**
+     * Used to use an expression as field value. Can be any expression
```
jmikola: "Allows any expression to be used as a field value."

lib/Doctrine/MongoDB/Aggregation/Builder.php
alcaeus: Didn't know MongoDB optimizes this. The opportunity presented itself by overwriting the ```sort()``` method within the ```Sort``` stage, that's where this optimization happens. I can refactor though to leave it out - this would create two sort and limit stages though.

lib/Doctrine/MongoDB/Aggregation/Expr.php
jmikola: Just to confirm: this is the catch-all for handling custom expressions that we don't have builder methods for, correct?

alcaeus: Yes - this allows you to toss in a nested expression object, e.g. one created with ```Builder::expr()``` in order to construct nested operators.

lib/Doctrine/MongoDB/Aggregation/Expr.php
```diff
@@ -0,0 +1,890 @@
...
+     * Defines an operator and value on the expression.
+     *
+     * If there is a current field, the operator will be set on it; otherwise,
+     * the operator is set at the top level of the query.
+     *
+     * @param string $operator
+     * @param array|self[]|self $expression
+     * @return self
+     */
+    protected function operator($operator, $expression)
```
jmikola: `operator()` is public in the basic query builder. Would making this public remove the need for the `expression()` method you have above? That would allow ultimate flexibility if the aggregate framework added more top-level operators (like `$and`), which do not require a field.

lib/Doctrine/MongoDB/Aggregation/Expr.php
jmikola: This was copied from the basic query builder method. Ideally, I think we could have `requiresCurrentField()` take an optional descriptive parameter to include in the exception message. As-is, the user needs to go through the backtrace one level to see the method that actually requires `field()`, since we throw from this vaguely named method.

Perhaps we can change `requiresCurrentField()`'s signature to `requiresCurrentField($method = null)` and then use `($method ?: 'This method')` when building the message? If we change it here, I think we can update the basic query builder as well in this same PR (changing both methods in a single commit).

lib/Doctrine/MongoDB/Aggregation/Expr.php
```diff
@@ -0,0 +1,890 @@
...
+    }
+
+    /**
+     * Ensure that a current field has been set.
+     *
+     * @throws LogicException if a current field has not been set
+     */
+    protected function requiresCurrentField()
+    {
+        if (!$this->currentField) {
```
jmikola: Spaces around `!` please :)

lib/Doctrine/MongoDB/Aggregation/Stage/GeoNear.php
```diff
@@ -0,0 +1,234 @@
...
+
+    /**
+     * @var boolean
+     */
+    protected $spherical = false;
+
+    /**
+     * @var boolean
+     */
+    protected $uniqueDocs;
```
jmikola: Any reason not to default to `private` for these, and other, properties?

lib/Doctrine/MongoDB/Aggregation/Stage/Match.php
jmikola: I think this would need to point to the `Query\Builder` class, so perhaps add a `use ... as QueryBuilder` at the top of this file, as was done previously with `QueryExpr`.

lib/Doctrine/MongoDB/Aggregation/Expr.php
alcaeus: :+1: on that. I'll change it to the same in ```Query::expr()``` as well.

lib/Doctrine/MongoDB/Aggregation/Stage/Match.php
```diff
@@ -0,0 +1,604 @@
...
+     *
+     * @see Expr::addAnd()
+     * @see http://docs.mongodb.org/manual/reference/operator/and/
+     * @param array|Expr $expression
+     * @return self
+     */
+    public function addAnd($expression)
```
jmikola: At a glance, it looks like this class is proxying all of the methods on the query builder. Would it make more sense to define a simple `__call()` magic method that allows a whitelist of method names to be proxied (passing all received arguments along in the process)? I suggest a whitelist because there are probably certain methods on the query builder that we don't want exposed (such as execution).

lib/Doctrine/MongoDB/Aggregation/Stage/Match.php
```diff
@@ -0,0 +1,604 @@
...
+     * The $language option may be set with {@link Builder::language()}.
+     *
+     * @see Expr::text()
+     * @see http://docs.mongodb.org/master/reference/operator/query/text/
+     * @param string $search
+     * @return self
+     */
+    public function text($search)
+    {
+        $this->query->text($search);
```
jmikola: FYI: This usage is limited a `$match` stage appearing first in the pipeline (see: [here](http://docs.mongodb.org/manual/tutorial/text-search-in-aggregation/)). I assume we can let MongoDB enforce this, though.

lib/Doctrine/MongoDB/Aggregation/Stage/Operator.php
```diff
@@ -0,0 +1,988 @@
...
+     * The arguments can be any valid expression as long as they resolve to either all numbers or to numbers and a date.
+     *
+     * @see http://docs.mongodb.org/manual/reference/operator/aggregation/add/
+     * @see Expr::add
+     * @param mixed|Expr $expression1
+     * @param mixed|Expr $expression2
+     * @param mixed|Expr $expression3,... Additional expressions
+     * @return self
+     */
+    public function add($expression1, $expression2 /* , $expression3, ... */)
```
jmikola: Same general comment as https://github.com/doctrine/mongodb/pull/213/files#r34468474. Does it make sense to replace most of the code in this class with a `__call()` for maintainability? I suppose the downside is that it's going to be unpleasant for folks expecting auto-completion in an IDE?

lib/Doctrine/MongoDB/Aggregation/Stage/Project.php
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    /**
+     * Shorthand method to exclude the _id field.
+     * @param bool $exclude
+     * @return self
+     */
+    public function excludeIdField($exclude = true)
+    {
+        return $this->field('_id')->expression(!$exclude);
```
jmikola: Spaces around ` ! `.

lib/Doctrine/MongoDB/Aggregation/Stage/Match.php
alcaeus: Yep, that was the thought. I would, however, point this out in the docblock, just like with ```$geoNear``` being the first stage in the pipeline.

lib/Doctrine/MongoDB/Aggregation/Stage/Project.php
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+namespace Doctrine\MongoDB\Aggregation\Stage;
+
+/**
+ * Fluent interface for adding a $project stage to an aggregation pipeline.
+ *
+ * @author alcaeus <alcaeus@alcaeus.org>
+ */
+class Project extends Operator
```
jmikola: Will this support projection beyond inclusion/exclusion of fields? I just want to make sure that we can still reshape documents (see: [here](http://docs.mongodb.org/manual/reference/operator/aggregation/project/#pipe._S_project)). Perhaps this is achieved by extending the Operator class, but that includes other things that don't quite make sense, like `addAnd()`. Do we need a more basic abstract class in that hierarchy that Project and Group can extend (both require expressions via fields at their top-level)?

lib/Doctrine/MongoDB/Aggregation/Stage/Redact.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+namespace Doctrine\MongoDB\Aggregation\Stage;
+
+/**
+ * Fluent interface for adding a $redact stage to an aggregation pipeline.
+ *
+ * @author alcaeus <alcaeus@alcaeus.org>
+ */
+class Redact extends Operator
```
jmikola: I believe Operator is too broad for Redact as well. A `$redact` stage is basically one expression that you might find on a single field, elsewhere. I don't think it makes sense to support things like `$and`, `$or`, and `$not`, since those would never resolve to an acceptable value:

> The argument can be any valid expression as long as it resolves to `$$DESCEND`, `$$PRUNE`, or `$$KEEP` system variables.

Source: http://docs.mongodb.org/manual/reference/operator/aggregation/redact/

tests/bootstrap.php
```diff
@@ -4,6 +4,6 @@
     throw new RuntimeException('Install dependencies to run test suite.');
 }
 
-$loader = require_once $file;
+$loader = require $file;
```
jmikola: Why the change? Was something including `bootstrap.php` twice?

lib/Doctrine/MongoDB/Aggregation/Builder.php
jmikola: > this would create separate sort and limit stages though

Not sure I follow. Don't we already create separate sort/limit stages?

lib/Doctrine/MongoDB/Aggregation/Stage/Match.php
jmikola: :+1:

lib/Doctrine/MongoDB/Aggregation/Builder.php
alcaeus: That was poor wording on my part. I meant to say "this would create separate sort and limit stages when these methods are called multiple times".
TBH, at least for limit I think we probably shouldn't do any optimizations. What's the expected output when doing this:
```$builder->limit(5)->limit(10)```
As useless as it seems, maybe the programmer did only expect to receive 5 documents while the current logic would overwrite the previous 5 with the second limit stage containing 10 items, thus producing a different result.

As for ```$sort```: does MongoDB combine multiple subsequent ```$sort``` stages into a single stage or does it run them one by one? In the latter case I'd rather allow for optimization in the builder to allow for subsequent calls to ```sort()``` within a loop or so.

File: lib/Doctrine/MongoDB/Aggregation/Stage/Operator.php (hunk 18 above)
Comment (alcaeus) [inline lib/Doctrine/MongoDB/Aggregation/Stage/Operator.php]: I mainly wrote this code this way because that's how it was done in ```Query\Builder``` and ```Query\Expr```. To be honest it's painstaking to write, it's a bunch of code duplication and we should actually have an interface that defines those methods to make the classes compatible (in the end it doesn't matter if I call ```add()``` on an ```Expr``` object or the ```Builder``` object itself.
As for code completion, this could be fixed by adding ```@method``` annotations to the class docblock. I know that PHPStorm parses them and offers code completion, don't know about the other IDEs though.

File: lib/Doctrine/MongoDB/Aggregation/Stage/Project.php (hunk 20 above)
Comment (alcaeus) [inline lib/Doctrine/MongoDB/Aggregation/Stage/Project.php]: Absolutely. The ```Operator``` class defines all stage operators that are common to ```$group```, ```$redact``` and ```$project```. This can be used to reshape documents as well, as you can see in the example test: https://github.com/doctrine/mongodb/pull/213/files#diff-41873c0811e1fca2085b746a594da896R107.
As for ```addAnd()``` and ```addOr()```, they absolutely make sense in that class and can be used in $project stages as well:
```
{
    $project: {
        someField: { $and: [ '$field', '$otherField' ] }
    }
}
```

However, due to a bug in my brain this is not possible since the ```$and``` and ```$or``` operators don't respect the current field - I'll fix this along with the other issues.

File: tests/bootstrap.php (hunk 22 above)
Comment (alcaeus) [inline tests/bootstrap.php]: This slipped through when committing - without this I couldn't get my local test suite to run, didn't mean to commit this. Will be removed ;)

File: lib/Doctrine/MongoDB/Aggregation/Stage/Redact.php (hunk 21 above)
Comment (alcaeus) [inline lib/Doctrine/MongoDB/Aggregation/Stage/Redact.php]: You're right - ```$redact``` is a bit of a special case. Technically, ```$cond``` and ```$ifNull``` would be enough on the root level, on the other hand maybe some user has a fancy document structure and handles redaction using a fancy ```$concat``` expression on his documents.
So, instead of going through every operator and thinking about "can the user actually use this in ```$redact``` I chose the lazy way and extended ```Operator```, thus leaving it to MongoDB to check whether the pipeline is a valid ```$redact``` or not. This only affects root level operators, as the user could do whatever he pleases within an ```$cond``` or ```$ifNull``` expression.

File: lib/Doctrine/MongoDB/Aggregation/Stage/Match.php (hunk 5 above)
Comment (alcaeus) [inline lib/Doctrine/MongoDB/Aggregation/Stage/Match.php]: Neither one. A new ```QueryExpr``` can be retrieved from the ```Builder```, but the method is not named ```expr()``` but ```matchExpr()```. The docblock was simply wrong. Speaking of which - does this create confusion? The builder has a ```expr()``` and ```matchExpr()``` method which return two different expression objects for different purposes.

File: lib/Doctrine/MongoDB/Aggregation/Expr.php (hunk 13 above)
Comment (alcaeus) [inline lib/Doctrine/MongoDB/Aggregation/Expr.php]: I think making this method public would expose too much functionality to the end user. The aggregation builder is meant to guide the user when writing aggregation queries, similar to the query builder. Allowing a "just put whatever you want in here" is a bit dangerous if you ask me.
The ```expression()``` method was added to allow creating creating nested documents in a ```$group``` or ```$project``` stage. That's also the reason why it requires ```field()``` to be called first.

Also, while the query builder uses the field for querying purposes, it's important to remember that most operators can be called without a field. The ```field()``` method defines the field where the result of an expression will be stored. Take this example:
```
$builder
    ->project()
    ->field('totalAmount')
    ->multiply('$taxRate', $this->expr()->subtract('$price', '$rebate'));
```
In this case, one operator is applied with a previous call to ```field()``` while the other operator is applied without calling ```field()``` because we don't need to store its value, we're just using it in the other expression.

File: lib/Doctrine/MongoDB/Aggregation/Builder.php (hunk 9 above)
Comment (jmikola) [inline lib/Doctrine/MongoDB/Aggregation/Builder.php]: [Aggregation Pipeline Optimization](http://docs.mongodb.org/manual/core/aggregation-pipeline-optimization/) explains some of the optimizations that MongoDB does for pipelines, but there may be more. The limit + limit coalescing is mentioned there. I'm generally in favor of simply letting the user construct exactly the pipeline they expect and letting the server do its optimization.

Additionally, you can always use the [`explain` option](http://docs.mongodb.org/manual/reference/command/aggregate/#return-information-on-the-aggregation-operation) with aggregation to see what some of the applied optimizations are.

File: lib/Doctrine/MongoDB/Aggregation/Stage/Operator.php (hunk 18 above)
Comment (jmikola) [inline lib/Doctrine/MongoDB/Aggregation/Stage/Operator.php]: I believe you mentioned in IRC that `@method` would have shortcomings for some of our variadic methods and users not being able to get the appropriate error or IDE feedback. It makes sense to keep things as they are and wait until we can make use of traits down the line.

File: lib/Doctrine/MongoDB/Aggregation/Stage/Match.php (hunk 16 above)
Comment (jmikola) [inline lib/Doctrine/MongoDB/Aggregation/Stage/Match.php]: Per https://github.com/doctrine/mongodb/pull/213#discussion_r34703338, we can disregard this.

File: lib/Doctrine/MongoDB/Aggregation/Stage/Operator.php (hunk 18 above)
Comment (alcaeus) [inline lib/Doctrine/MongoDB/Aggregation/Stage/Operator.php]: Yes - using ```@method``` and ```__call``` method would mean we'd have to implement checks to see if the correct number of arguments is given for each operator. This way we get those checks for free. I'll leave the code as is.

File: lib/Doctrine/MongoDB/Aggregation/Builder.php (hunk 9 above)
Comment (alcaeus) [inline lib/Doctrine/MongoDB/Aggregation/Builder.php]: Yeah, I think it's better to just let MongoDB do its thing. I'll remove the optimizations, probably saves a lot of troubles.

File: lib/Doctrine/MongoDB/Aggregation/Stage/Project.php (hunk 20 above)
Comment (jmikola) [inline lib/Doctrine/MongoDB/Aggregation/Stage/Project.php]: > this is not possible since the `$and` and `$or` operators don't respect the current field - I'll fix this along with the other issues.

What do you mean by that? I'd expect it to resolve to `{ $project: { someField: <true|false> }}`. This also shows up in [this example](http://docs.mongodb.org/manual/reference/operator/aggregation/and/#example) from the documentation.

File: lib/Doctrine/MongoDB/Aggregation/Stage/Redact.php (hunk 21 above)
Comment (jmikola) [inline lib/Doctrine/MongoDB/Aggregation/Stage/Redact.php]: :+1: I'm fine with leaving the database to catch any errors.

File: lib/Doctrine/MongoDB/Aggregation/Stage/Project.php (hunk 20 above)
Comment (alcaeus) [inline lib/Doctrine/MongoDB/Aggregation/Stage/Project.php]: Let me rephrase that: "This was not possible due to an issue in the code that has since been fixed."

File: lib/Doctrine/MongoDB/Aggregation/Stage/Redact.php (hunk 21 above)
Comment (alcaeus) [inline lib/Doctrine/MongoDB/Aggregation/Stage/Redact.php]: :+1:

**Comment by jmikola** [comment]
@alcaeus: I restarted the two hung builds earlier today and it looks like they've completed. Are we good to go here?

**Comment by alcaeus** [comment]
Yep - good to go. I thought about moving the ```field()``` method out of ```Expr``` and into the ```$group```, ```$project``` and ```$redact``` stages separately since an Expression doesn't usually need the functionality with a field name, only root level expressions in an aggregation stage need it. It would be quite a change but clean up the API. What do you think?

**Comment by jmikola** [comment]
> It would be quite a change but clean up the API. What do you think?

Sounds reasonable to me. It will also make it less likely for a user to shoot themselves in the foot by generating an invalid pipeline.

**Comment by jmikola** [comment]
> since an Expression doesn't usually need the functionality with a field name

`$project` might need special consideration here. See: [Embedded Document Fields](http://docs.mongodb.org/manual/reference/operator/aggregation/project/#embedded-document-fields), where this is valid projection on `contact`:

```
contact: { address: { country: <1 or 0 or expression> } }
```

**Comment by alcaeus** [comment]
> ```$project``` might need special consideration here.

Yes, ```$project``` and ```$group``` both need an ```Expr``` class that has a ```field()``` method to support embedded documents. In that case I suggest we leave it as it is now to avoid having to introduce another ```Expr``` class in addition to the two classes we already have (```Aggregation\Expr``` and ```Query\Expr```).
This might leave some options where users might shoot themselves in the foot but I don't think those cases are worth introducing a bunch of classes.

**Comment by jmikola** [comment]
LGTM. Feel free to do the honors :)

**Comment by metanav** [comment]
Glad to see that this feature is available in the version 1.0.1.
It would be more useful if hydration can be enabled for those queries which return single or collection of the same document on which aggregation is run.

**Comment by malarzm** [comment]
@metanav as far as I know @alcaeus has plans to improve aggregation for ODM, this is just the base :)

**Comment by alcaeus** [comment]
As @malarzm has stated, this is just the base required to support hydrated aggregation results in ODM. Hydrated aggregation results are a bit complicated since it requires a lightweight way to map results - every aggregation query needs a different result document, unlike regular documents where every query produces the same result. I'll pick this up in the coming weeks - look for a PR in https://github.com/doctrine/mongodb-odm ;)

**Comment by metanav** [comment]
Sometimes it is needed to pass already built query (using QueryBuilder or manuallly) to the AggregationBuilder match stage as an array (or object, I don't know if both builders are compatible to each other). Is there any way to achieve this in the current version or any possibility to add this feature to make it more flexible?

**Comment by alcaeus** [comment]
You mean something like ```$queryBuilder->matchArray(['foo' => 'bar']);```? In that case, no, it's not possible and when thinking of the API I didn't deem it necessary. If you could explain your use case I could give it some thought and see if we can come up with something.

**Comment by metanav** [comment]
I do not think it is not possible, may be I could not explain it.
I've created an example gist here:  https://gist.github.com/metanav/ace4e5de7aecd556c5a8

**Comment by alcaeus** [comment]
Gotcha. What you can do is abuse the fact that using ```equals``` without a ```currentField``` sets your query to whatever you pass it. In your case:
```
if (is_array($filter)) {
    $ab->match()->equals($filter);
} else {
    $ab->match()->text($filter);
}
```

I tried it by adding more fields after that as well:
```
public function testFilterArray()
{
    $filter = ['name' => 'keyword'];

    $matchStage = new Match($this->getTestAggregationBuilder());
    $matchStage
        ->equals($filter)
        ->field('someField')
        ->equals('someValue');

    $expectedMatch = ['name' => 'keyword', 'someField' => 'someValue'];
    $this->assertEquals(['$match' => $expectedMatch], $matchStage->getExpression());
}
```

metanav:
Thanks, worked like a charm!  I hope this is not an artefact and should not be lost in some fixes :).

alcaeus:
It is a feature of the query builder (on which the match stage relies) and has been there for some time. I think you might be safe ;)

metanav:
I guess DateTime object to MongoDate object conversion is not being done automatically as we have in Query builder.

alcaeus:
I'm guessing you're thinking of ODM. This library sits between the driver and ODM and just provides nicer interfaces for the legacy driver. Type conversion is handled only by the ODM - you'll have to do your own conversions.

metanav:
Oh yes, got it!

alcaeus:
FWIW, I should be able to get back to doctrine/mongodb-odm#1263 soon now that we've gotten rid of old PHP and driver versions across the board. That library will do type conversion, at least in early ```$match``` stages. Not sure how much type conversion is going to be possible once projection and grouping kicks in - we don't have any mapping information for stages in the middle of the pipeline.

metanav:
Would priming be possible in doctrine/mongodb-odm#1263 at least in $match stages?

alcaeus:
What exactly do you mean? Priming has nothing to do with individual stages, just result. Let's say your aggregation result document contains a reference to another document: this is where priming is effective since it fetches all referenced documents at once. However, this doesn't affect the execution of the pipeline itself, which is done by MongoDB. This library provides a clean interface to create pipelines, ODM will give you type conversion and hydration into nice objects instead of dealing with arrays. Everything else is MongoDB and the driver doing their things.

metanav:
Yes, priming has nothing to do with match. I should have  asked if priming is possible in hydration?

alcaeus:
I'll certainly put it on the list of things that could be useful. Not sure I can add it right from the get-go.

PR #215, opened 2015-05-16
Test with lowest dependencies.

BernardoSilva:
Run library against `lowest` dependencies using `mongo stable` version.
This will find out if lowest dependencies need to be updated.

 - [x] composer is now updated on every build.
 - [x] Update `doctrine/common` min dependency to `~2.2`.

Can check build failing for 2.1 [here](https://travis-ci.org/amyboyd/mongrate/jobs/62818098#L124)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,5 @@ services: mongodb
 before_script:
   - yes '' | pecl -q install -f mongo-${MONGO_VERSION}
   - php --ri mongo
+  - composer self-update
   - composer update --dev --no-interaction --prefer-source $PREFER_LOWEST
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,15 +7,16 @@ php:
   - 5.6
 
 env:
-  - MONGO_VERSION=1.2.12
-  - MONGO_VERSION=1.3.2
-  - MONGO_VERSION=1.3.7
-  - MONGO_VERSION=1.4.5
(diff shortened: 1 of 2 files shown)


jmikola:
The library actually appears to be compatible with Doctrine Common 2.1.x (I just ran tests locally and the classes we depend on are there); however, the issue is that those early releases did not include an `autoload` configuration block for Composer, which meant that PHP could never load the reference classes. These commits, which are first available in 2.2.0, appeared to fix it:

 * https://github.com/doctrine/common/commit/6a761386a20ba91a16835992fd8b5e70e1efb716
 * https://github.com/doctrine/common/commit/1f9022833313bc5e53749ad497fd4e24ea701988
 * https://github.com/doctrine/common/commit/b44ea1dc14e9bbb1a39b37a39875f7a8d4351b82

I don't see any reason we can't bump our dep, especially since even 2.2 is quite old and this is just going into the 1.2.x branch. Thanks for the contribution!

jmikola:
Manually merged in https://github.com/doctrine/mongodb/commit/c3c029cab1917afb8fa7cc41ef3dc328e8a852fb.

BernardoSilva:
Cool :+1: 
I thought I saw the dependencies in `2.1` I just did not understand why they were failing, so I used `2.2` to ensure builds were still passing and everything was working properly.

PR #216, opened 2015-05-16
Remove composer.lock to avoid locking dependencies.

BernardoSilva:
I don't see any advantage of leaving the lock file in here.
This is causing dependencies to be locked to a specific revision.
This would be good if this was using unstable dependencies where you want to lock to a specific commit.

Locking to `v2.4.2`
```sh
- Installing doctrine/common (v2.4.2)
    Downloading: 100% 
```


jmikola:
See @naderman's comment in https://github.com/doctrine/mongodb/pull/207#issuecomment-102647182.

BernardoSilva:
:+1:

PR #218, opened 2015-05-31
Change parameter for $currentDate operator

### LINKED ISSUE #217: Change Builder and Expr currentDate() method to accept string $type
See https://github.com/doctrine/mongodb/pull/212/files#r31166431:

> What do you think about changing this to $type = 'date'? We should generally discourage users from using the BSON timestamp type, since it's actually more than a UNIX timestamp. A boolean option makes it very easy for them to inadvertently use it. This also keeps the method flexible if the $currentDate operator changes down the line to support additional types (e.g. date with a timezone).

alcaeus:
This changes the $currentDate operator to no longer accept a timestamp parameter, instead accepting the type directly.

Fixes #217


jmikola:
LGTM!

PR #223, opened 2015-08-14
Support $useKeys option when EagerCursor converts to an array

### LINKED ISSUE #160: EagerCursor does not allow objects as identifiers (eg. BSON BinData)
There currently is no way of correctly using an eager cursor if your _id is of a complex datatype. The reason is because the toArray() implementation uses the _id as keys, forcing the keys to be typecast and overriding eachother.

I believe the solution lies in EagerCursor to accept an additional $useKeys argument in the constructor to be used when calling EagerCursor::initialize(). The second part is more open to discussion however, how should this argument be exposed to the query builder API? A possible solution would be to change the signature of the Builder::eagerCursor() method to accept $useKeys as an optional second argument and using this option in prepareCursor().

jmikola:
Fixes #160.

### DIFF
--- a/lib/Doctrine/MongoDB/EagerCursor.php
+++ b/lib/Doctrine/MongoDB/EagerCursor.php
@@ -50,13 +50,27 @@ class EagerCursor implements CursorInterface
     protected $initialized = false;
 
     /**
+     * Whether to preserve keys from the Cursor (i.e. "_id" values) when
+     * initializing the $data array.
+     *
+     * @var boolean
+     */
+    protected $useKeys = true;
+
+    /**
      * Constructor.
      *
+     * If documents in the result set use BSON objects for their "_id", the
+     * $useKeys parameter may be set to false to avoid errors attempting to cast
+     * arrays (i.e. BSON objects) to string keys.
+     *
      * @param Cursor $cursor
+     * @param boolean $useKeys
      */
-    public function __construct(Cursor $cursor)
+    public function __construct(Cursor $cursor, $useKeys = true)
     {
         $this->cursor = $cursor;
+        $this->useKeys = (boolean) $useKeys;
     }
 
     /**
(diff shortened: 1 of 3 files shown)


jmikola:
@alcaeus: Can you take a look at this Monday?

alcaeus:
The changes look good :+1: 
Do we want to expose this in the ```Builder``` or ```Query``` class so that it can be controlled by developers?

jmikola:
> Do we want to expose this in the ```Builder``` or ```Query``` class so that it can be controlled by developers? 

@alcaeus: Possibly. At the moment, it appears the only way to specify `$useKeys` would be to let the query builder return a cursor and then invoke `toArray()` on it manually. The Query class' own `toArray()` method doesn't currently pass through the option. `getSingleResult()` does so, but that's simply because `Cursor::getSingleResult()` always sets it on its own.

This may be because we're being strict about Doctrine MongoDB's Iterator and IteratorAggregate interfaces, neither of whose `toArray()` methods take an argument.

lib/Doctrine/MongoDB/EagerCursor.php
```diff
@@ -107,7 +121,7 @@ public function getSingleResult()
     public function initialize()
     {
         if ($this->initialized === false) {
-            $this->data = $this->cursor->toArray();
+            $this->data = $this->cursor->toArray($this->useKeys);
```
jmikola: Going to change this to just have a `$useKeys` param on `initialize()`. We don't need it on the constructor.

jmikola: Although we discussed moving `$useKeys` to an `initialize()` parameter, integration with ODM would benefit from keeping this as a constructor argument. That is because ODM's Cursor class (if we remove its EagerCursor) will merely wrap a CursorInterface.

PR #224, opened 2015-08-14
Support client and socket timeout options in Collection::count()

### LINKED ISSUE #221: timeout/maxTimeMS settings for count()
Current implementation of \Doctrine\MongoDB\Collection::count() doesn't support options passing to http://php.net/manual/en/mongocollection.count.php

Hence its impossible to set custom timeout for count() queries except by using \MongoCursor::$timeout - but that's deprecated.

After deprecation of \MongoCursor::$timeout the whole timeout thing became very inconsistent - its suggested to use $cursor->timeout() but that works only for queries that return Cursor. I.e. for map-reduce queries an array is returned instead and hence ->timeout() thing won't work. In that case the only option is to set timeout during query building, e.g. $queryBuilder->getQuery(['timeout' => 60000])->execute(). I think there should be a single and straightforward way to pass options / set timeout etc. for any kind of query - that's what abstraction layer is for.

jmikola:
Fixes #221.

### DIFF
--- a/lib/Doctrine/MongoDB/Collection.php
+++ b/lib/Doctrine/MongoDB/Collection.php
@@ -166,17 +166,21 @@ class Collection
      *
      * @see http://php.net/manual/en/mongocollection.count.php
      * @see http://docs.mongodb.org/manual/reference/command/count/
-     * @param array   $query
-     * @param integer $limit
-     * @param integer $skip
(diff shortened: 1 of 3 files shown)


alcaeus:
Merged manually in 484e907.

PR #225, opened 2015-08-16
Include all cursor methods in cursor interface

alcaeus:
This extends the common cursor interface introduced in #209 to include all methods currently available in the cursor. This makes the ```Cursor``` and ```EagerCursor``` more interchangeable in ODM, allowing us to plug both classes into ```ODM\Cursor``` once that has been adapted.

The ```hasNext``` and ```getNext``` methods have to be simulated by advancing and rewinding the array pointer as well as checking if the cursor has started iterating.

### DIFF
--- a/lib/Doctrine/MongoDB/CursorInterface.php
+++ b/lib/Doctrine/MongoDB/CursorInterface.php
@@ -19,12 +19,10 @@
 
 namespace Doctrine\MongoDB;
 
-use Doctrine\MongoDB\Util\ReadPreference;
-
 /**
  * Wrapper for the PHP MongoCursor class.
  *
- * @since  1.0
+ * @since  1.2
  * @author alcaeus <alcaeus@alcaeus.org>
  */
 interface CursorInterface extends Iterator
@@ -76,10 +74,41 @@ interface CursorInterface extends Iterator
     /**
      * Wrapper method for MongoCursor::fields().
      *
+     * @param array $f Fields to return (or not return).
+     *
      * @see http://php.net/manual/en/mongocursor.fields.php
      * @return self
      */
     public function fields(array $f);
+
+    /**
+     * Return the collection for this cursor.
+     *
+     * @return Collection
+     */
+    public function getCollection();
+
+    /**
+     * Return the selected fields (projection).
+     *
+     * @return array
+     */
+    public function getFields();
+
+    /**
+     * Wrapper method for MongoCursor::getNext().
+     *
+     * @see http://php.net/manual/en/mongocursor.getnext.php
+     * @return array|null
+     */
+    public function getNext();
+
+    /**
+     * Return the query criteria.
+     *
+     * @return array
+     */
+    public function getQuery();
 
     /**
      * Wrapper method for MongoCursor::getReadPreference().
@@ -98,6 +127,14 @@ interface CursorInterface extends Iterator
      * @return self
      */
     public function setReadPreference($readPreference, array $tags = null);
+
+    /**
+     * Wrapper method for MongoCursor::hasNext().
+     *
+     * @see http://php.net/manual/en/mongocursor.hasnext.php
+     * @return boolean
+     */
+    public function hasNext();
 
     /**
      * Wrapper method for MongoCursor::hint().
@@ -133,6 +170,11 @@ interface CursorInterface extends Iterator
      * @return self
      */
     public function limit($num);
+
+    /**
+     * Recreates the internal MongoCursor.
+     */
+    public function recreate();
 
     /**
      * Wrapper method for MongoCursor::reset().
--- a/lib/Doctrine/MongoDB/CursorInterface.php
+++ b/lib/Doctrine/MongoDB/CursorInterface.php
@@ -185,6 +185,15 @@ interface CursorInterface extends Iterator
     public function reset();
 
     /**
+     * Set whether to use the document's "_id" value as its iteration key.
+     *
+     * @since 1.2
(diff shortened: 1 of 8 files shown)


malarzm:
:+1:

lib/Doctrine/MongoDB/CursorInterface.php
```diff
@@ -76,12 +74,51 @@ public function explain();
...
+     * Return the collection for this cursor.
+     *
+     * @return Collection
+     */
+    public function getCollection();
+
+    /**
+     * Return the connection for this cursor.
+     *
+     * @deprecated 1.1 Will be removed for 2.0
```
jmikola: If this is deprecated, so we really want to add it to the interface?

lib/Doctrine/MongoDB/CursorInterface.php
```diff
@@ -100,6 +137,25 @@ public function getReadPreference();
...
     /**
+     * Reset the cursor and return its first result.
+     *
+     * The cursor will be reset both before and after the single result is
+     * fetched. The original cursor limit (if any) will remain in place.
+     *
+     * @see Iterator::getSingleResult()
+     * @return array|object|null
+     */
+    public function getSingleResult();
```
jmikola: Is this redundant if Iterator is already defining the method?

lib/Doctrine/MongoDB/EagerCursor.php
```diff
@@ -50,6 +50,13 @@ class EagerCursor implements CursorInterface
     protected $initialized = false;
 
     /**
+     * Whether the cursor started iterating. This is necessary to get getNext() and hasNext() to work properly
```
jmikola: I would put "This is needed to implement getNext() and hasNext()" as its own paragraph (after two line breaks).

lib/Doctrine/MongoDB/EagerCursor.php
```diff
@@ -50,6 +50,13 @@ class EagerCursor implements CursorInterface
...
+     *
+     * @var bool
```
jmikola: I think we use "boolean" in most places?

lib/Doctrine/MongoDB/EagerCursor.php
```diff
@@ -219,6 +230,54 @@ public function fields(array $f)
...
+        return $this->cursor->getCollection();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    public function getConnection()
+    {
+        return $this->cursor->getConnection();
+    }
```
jmikola: If we remove this method from the interface (because it's deprecated, I suppose we don't need to add it here); however, if you'd like to keep it in the interface, by all means leave it.

lib/Doctrine/MongoDB/EagerCursor.php
```diff
@@ -219,6 +230,54 @@ public function fields(array $f)
...
+    {
+        $this->initialize();
+
+        $next = ($this->iterating) ? next($this->data) : current($this->data);
+        $this->iterating = true;
+        if ($next !== false) {
+            return $next;
+        }
+
+        return null;
```
jmikola: `return $next !== false ? $next : null` works, too.

lib/Doctrine/MongoDB/EagerCursor.php
```diff
@@ -237,6 +296,26 @@ public function setReadPreference($readPreference, array $tags = null)
...
+        $this->initialize();
+
+        $wasIterating = $this->iterating;
+        $hasNext = $this->getNext() !== null;
+
+        // Reset the internal cursor if we weren't iterating
+        if ($wasIterating) {
+            prev($this->data);
+        } else {
+            $this->iterating = false;
```
jmikola: Is it purposeful that you want to leave the cursor in a "non-iterating" state, while still being initialized?

lib/Doctrine/MongoDB/CursorInterface.php
alcaeus: I was thinking no, until I saw that ```ODM\Cursor``` exposes the method as well (where I wanted to use the same interface). If we leave it out of ODM we can remove it from the interface here, otherwise I'd leave it in. What's your opinion?

lib/Doctrine/MongoDB/EagerCursor.php
alcaeus: ```iterating``` has to be reset to false, otherwise calling ```hasNext()``` followed by ```getNext()``` will return the second result instead of the first result as would be expected.

If the cursor is marked as "iterating", ```getNext()``` will cause the ```next``` function which will advance the array pointer to the second result and return this. The first result however has never been fetched and would be skipped.

lib/Doctrine/MongoDB/CursorInterface.php
jmikola: As discussed, we can remove this from the interface and have ODM still provide the method in its Cursor class, but simply do `$this->baseBursor->getCollection()->getDatabase()->getConnection();`

lib/Doctrine/MongoDB/EagerCursor.php
jmikola: This method can be removed entirely, since we're cutting it from the interface.

jmikola:
LGTM.

lib/Doctrine/MongoDB/EagerCursor.php
jmikola: :+1:

PR #227, opened 2015-08-19
Specify time limit operation on a mongodb cursor

ne0h12:
Termination processing operation if the associated cursor exceeds its allotted time limit. It is beneficial when PHP driver goes on timeout (by default 30 seconds). As a results mongodb keep health on highload projects because of not performing not needed operations.

Thanks. Sorry for my english.

### DIFF
--- a/lib/Doctrine/MongoDB/Cursor.php
+++ b/lib/Doctrine/MongoDB/Cursor.php
@@ -78,6 +78,7 @@ class Cursor implements CursorInterface
     protected $sort;
     protected $tailable;
     protected $timeout;
+    protected $maxTimeMS;
 
     /**
      * Constructor.
@@ -661,6 +662,22 @@ class Cursor implements CursorInterface
     {
         $this->timeout = (integer) $ms;
         $this->mongoCursor->timeout($ms);
+        return $this;
+    }
+
+    /**
+     * Wrapper method for MongoCursor::maxTimeMS().
+     *
+     * @see http://php.net/manual/en/mongocursor.maxtimems.php
+     * @param integer $ms
+     * @return self
+     */
+    public function maxTimeMS($ms)
+    {
+        if (version_compare(phpversion('mongo'), '1.5.0', '>=')) {
+            $this->maxTimeMS = (integer)$ms;
+            $this->mongoCursor->maxTimeMS((integer)$ms);
+        }
         return $this;
     }
 
--- a/lib/Doctrine/MongoDB/Cursor.php
+++ b/lib/Doctrine/MongoDB/Cursor.php
@@ -509,6 +509,9 @@ class Cursor implements CursorInterface
         if ($this->timeout !== null) {
             $this->mongoCursor->timeout($this->timeout);
         }
+        if ($this->maxTimeMS !== null) {
+            $this->mongoCursor->maxTimeMS($this->maxTimeMS);
+        }
     }
 
     /**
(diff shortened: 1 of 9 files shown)


lib/Doctrine/MongoDB/CursorInterface.php
```diff
@@ -253,4 +253,13 @@ public function tailable($tail = true);
...
     public function timeout($ms);
+
+    /**
+     * Wrapper method for MongoCursor::maxTimeMS().
+     *
+     * @see http://php.net/manual/en/mongocursor.maxtimems.php
+     * @param integer $ms
+     * @return self
+     */
+    public function maxTimeMS($ms);
```
jmikola: @alcaeus: We may have shot ourselves in the foot here, but AFAIK we cannot add a new method to the interface without bumping to a new major version. This has come up with the collection interfaces in Doctrine Common before.

/cc @Ocramius

alcaeus: Technically speaking, yes. This is a BC break and thus would force us to go 2.0. I'm kicking myself now for not catching all methods that were added in the 1.5 driver.

ne0h12:
@jmikola why Scrutinizer says that checks failed, but when I opening details, then I do not see the errors ? What am i doing wrong ? Thank.

jmikola:
@ne0h12: I'm not sure what happened with Scrutinizer, but I wouldn't worry too much about it.

lib/Doctrine/MongoDB/Cursor.php
```diff
@@ -665,6 +669,22 @@ public function timeout($ms)
...
+     * @param integer $ms
+     * @return self
+     */
+    public function maxTimeMS($ms)
+    {
+        if (version_compare(phpversion('mongo'), '1.5.0', '>=')) {
+            $this->maxTimeMS = (integer)$ms;
+            $this->mongoCursor->maxTimeMS((integer)$ms);
+        }
+        return $this;
```
jmikola: @ne0h12: The most compatible solution would be to call [`MongoCursor::addOption('$maxTimeMS', $ms)`](http://us2.php.net/manual/en/mongocursor.addoption.php) here.

lib/Doctrine/MongoDB/CursorInterface.php
jmikola: @ne0h12: Unfortunately, this needs to be removed. We cannot add new interface methods without bumping a major version.

tests/Doctrine/MongoDB/Tests/CursorTest.php
```diff
@@ -381,6 +381,44 @@ public function testRecreate()
         $cursor->recreate();
     }
 
+    public function testSetMaxTimeMSWhenRecreateCursor()
+    {
+        if (version_compare(phpversion('mongo'), '1.5.0', '<')) {
+            $this->markTestSkipped('This test is not applicable to driver versions < 1.5.0');
```
jmikola: If we use `addOption()`, I think we can run this test on any driver version with appropriate mock changes.

jmikola:
@alcaeus: Note, this also needs to go into master (it was rightfully opened against it) since we're adding new builder methods.

lib/Doctrine/MongoDB/CursorInterface.php
ne0h12: Ok, I removed it

alcaeus:
@jmikola I agree.

alcaeus:
Closed manually in https://github.com/doctrine/mongodb/commit/7b4d519bdf28289a49545ee804fb62ed99ddd1b2. Thanks!

PR #229, opened 2015-08-20
Fix EagerCursor::skip() method calling limit method of base cursor in…

munkie:
…stead of skip method


alcaeus:
Right. Thank you ;)

PR #231, opened 2015-08-24
Remove count method declaration in CursorInterface

### LINKED ISSUE #230: Fatal error on php: 5.3.6
Hello,

The error is: 
Fatal error: Can't inherit abstract function Countable::count() (previously declared abstract in Doctrine\MongoDB\CursorInterface) in doctrine/mongodb/lib/Doctrine/MongoDB/CursorInterface.php on line 29

This interface extends another 2 interfaces and in one of them the count method is again presented

alcaeus:
Fixes #230.


jmikola:
LGTM. I don't think we need the docs kept around, as this omits the `$foundOnly` option found on the driver method anyway.

PR #234, opened 2015-10-19
Add support for the $comment operator

alcaeus:
As mentioned in doctrine/mongodb-odm#1224, there is no support for the $coment operator that was introduced in 2.4. This PR fixes that.


malarzm:
:+1:

PR #235, opened 2015-10-31
Add support for $setOnInsert operator

alcaeus:
This PR adds support for the [$etOnInsert operator](https://docs.mongodb.org/manual/reference/operator/update/setOnInsert/) that was added in MongoDB 2.4.

### DIFF
--- a/tests/Doctrine/MongoDB/Tests/Query/BuilderTest.php
+++ b/tests/Doctrine/MongoDB/Tests/Query/BuilderTest.php
@@ -261,6 +261,26 @@ class BuilderTest extends \PHPUnit_Framework_TestCase
         $this->assertEquals($expected, $qb->getNewObj());
     }
 
+    public function testSetOnInsert()
+    {
+        $createDate = new \MongoDate();
+        $qb = $this->getTestQueryBuilder()
+            ->update()
+            ->upsert()
+            ->field('username')->equals('boo')
+            ->field('createDate')->setOnInsert($createDate);
+
+        $expected = array(
+            'username' => 'boo'
+        );
+        $this->assertEquals($expected, $qb->getQueryArray());
+
+        $expected = array('$setOnInsert' => array(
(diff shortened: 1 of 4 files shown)


tests/Doctrine/MongoDB/Tests/Query/BuilderTest.php
```diff
@@ -177,7 +177,6 @@ public function testFindQuery()
     public function testUpsertUpdateQuery()
     {
         $qb = $this->getTestQueryBuilder()
-            ->update()
```
malarzm: wasn't this test checking if `upsert` overrides previous `update`?

tests/Doctrine/MongoDB/Tests/Query/BuilderTest.php
```diff
@@ -261,6 +260,25 @@ public function testUnsetField()
...
+            ->field('username')->equals('boo')
+            ->field('createDate')->setOnInsert(new \MongoDate());
+
+        $expected = array(
+            'username' => 'boo'
+        );
+        $this->assertEquals($expected, $qb->getQueryArray());
+
+        $expected = array('$setOnInsert' => array(
+            'createDate' => new \MongoDate()
```
malarzm: Isn't this making test fragile? I think that storing `new \MongoDate()` in variable and using it in qb and here could be better

tests/Doctrine/MongoDB/Tests/Query/BuilderTest.php
alcaeus: You're correct. I've had half a mind of making upserts imply an update but then I changed my mind and forgot to change this back. Fixed now.

tests/Doctrine/MongoDB/Tests/Query/BuilderTest.php
alcaeus: You're absolutely right. Changed it.

PR #238, opened 2015-11-24
Bump PHP and mongo version requirements

alcaeus:
PHP 5.3 is no longer supported while PHP 5.4 is getting security updates only, so the next minor version upgrade (1.3.0) will require PHP 5.5 or 5.6. Please note that currently PHP 7 is not supported.
While we're at it, the new version will also bump the minimum driver versions. Thus, the 1.2, 1.3 and 1.4 versions of the mongo driver will no longer be supported and 1.5.0 or newer will be required.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,16 +2,11 @@ sudo: false
 language: php
 
 php:
-  - 5.3
-  - 5.4
   - 5.5
   - 5.6
 
 env:
-  - MONGO_VERSION=1.2.12 PREFER_LOWEST=""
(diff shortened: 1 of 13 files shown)


jmikola:
If we're doing this, there should be a fair bit of compatibility code eligible for removal. Try grepping the library for `version_compare()` and `method_exists()`.

alcaeus:
Thanks for the heads up, will do that! :+1:

jmikola:
LGTM :+1:

Issue #151, opened 2014-01-23
Better query logging capabilities


woodworker:
The logging capabilities in the mongodb Odm are not as good as they could be.

I would like to have similar functionality to the doctrine orm SQLLogger Interface
see https://github.com/doctrine/dbal/blob/master/lib/Doctrine/DBAL/Logging/SQLLogger.php

i could think of some ways to do this

1. a second parameter for a after log method
 * makes getloggerCallable incompatible and looksugly
2. a new method in Doctrine\MongoDB\Configuration
 * setAfterQueryLoggerCallable
 * not a good solution to add new config parameter for every new logging case
3. make setLoggerCallable incompatible and use a similar interface like Doctrine ORM SQLLogger
 * BC break

Any other ideas or comments?

PS: i would contribute a patch for this feature request when there is consensus for a solution

jmikola:
Closing this due to #153 also being closed. AFAIK, you were able to get by with driver's [stream context listeners](http://php.net/manual/en/mongo.context.php).

PR #245, opened 2016-01-09
Feature proposal: Add hooks around fetch single result from curosr

blockjon:
I'd like to track how long it takes to fetch an item across the db handle when I do a fetchSingleResult. Seems like this is a way to do it. Seeking comment.

### DIFF
--- a/lib/Doctrine/MongoDB/Collection.php
+++ b/lib/Doctrine/MongoDB/Collection.php
@@ -1511,4 +1511,13 @@ class Collection
             array_intersect_key($options, $keys),
         );
     }
+
+    /**
+     * Return the EventManager attached to this collection.
+     *
+     * @return EventManager
+     */
+    public function getEventManager() {
+        return $this->eventManager;
+    }
 }
--- a/lib/Doctrine/MongoDB/Cursor.php
+++ b/lib/Doctrine/MongoDB/Cursor.php
@@ -19,6 +19,9 @@
 
 namespace Doctrine\MongoDB;
 
+use Doctrine\Common\EventManager;
+use Doctrine\MongoDB\Event\FetchSingleResultFromCursorEventArgs;
+
 /**
  * Wrapper for the PHP MongoCursor class.
  *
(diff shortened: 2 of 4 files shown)


alcaeus:
To be honest, I'm not sure adding events purely for measuring time is something you should need:
- if you want to track how fast/slow PHP code is, use a profiler like Blackfire or Tideways.
- if you need to figure out which queries are run, you can add a logger to the configuration to have instances of `LoggableCollection` which logs every query to a logger of your choice.

I'm not sure what else those events could be useful for, maybe you could elaborate a bit?

alcaeus:
Just adding one more comment because I knew there were some logging things going around before the 1.2 release. If you're interested in logging that the driver and MongoDB itself are doing, @jmikola [previously recommended](https://github.com/doctrine/mongodb/issues/151#issuecomment-114252226) using [stream context listeners](https://secure.php.net/manual/en/mongo.context.php) in the driver.

blockjon:
I have a different perspective, perhaps because I'm used to simpler database like MySQL.

In MySQL frameworks, it is very easy to measure query performance from the perspective of PHP because there's a very clear moment when the query takes place.

With Doctrine ODM, the moment that that the query takes place is hidden within the Cursor class. The query is implicitly executed as a consequence of trying to convert the results to an array.

With our application, we are concerned with how long our queries are taking from the PHP perspective. I want to be able to add instrumentation using our current instrumentation framework.

It is my opinion that the moment a query fires is a crucial moment in the control flow of any application and we should have a very elegant and simple way to have hooks around that moment within the ODM. Tooks like XHProf, Blackfire, and Tideways are things which we've looked at to perform profiling and sampling of the live environment for brief periods of time, but we are looking for a solution to allow us to instrument query performance long term.

blockjon:
@alcaeus ^

blockjon:
The exact ticket I'm working on right now has on it:

"It would be great to be able to measure timing of different specific queries (for instance, timing of just a primary key query, versus of a scatter get, etc.)."

I'm not even sure how I would do that with with the ODM since the query is prepared in one area and its executed in another area.

Ideally, I would have code that *roughly* could do this:

```
$query = .....;
$startTimer = now;
$result = $query->execute();
$ms = $now - $startTimer;
if ($query->isScatterGet()) {
  $instrument->increment('suboptimalmongoscattergetquery', $ms);
  $log->info("Scatter get query detected: $query->toString() took $ms to execute.");
} else {
  $instrument->increment('optimalprimarykeyquery', $ms);
}
```

alcaeus:
Queries in doctrine/mongodb are just objects that hold a query array, an optional update array and some options. When you call `$query->execute()`, you get a cursor back, that's it. There is no explanation of the query attached to the cursor you're getting back, so you won't know whether that was an optimal or a suboptimal query.

As I've said before, the use case in question would best be covered logging these things either at the database level (check out the instrumentation provided by MongoDB itself, e.g. MMS) or at the driver level.

PR #246, opened 2016-01-13
Open up aggregation builder API for use in ODM

alcaeus:
This PR collects all changes necessary to support hydrated aggregation results in doctrine/mongodb-odm#1263.

### DIFF
--- a/lib/Doctrine/MongoDB/Aggregation/Builder.php
+++ b/lib/Doctrine/MongoDB/Aggregation/Builder.php
@@ -303,7 +303,7 @@ class Builder
      * @param Stage $stage
      * @return Stage
      */
-    private function addStage(Stage $stage)
+    protected function addStage(Stage $stage)
     {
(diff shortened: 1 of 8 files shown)


lib/Doctrine/MongoDB/Aggregation/Builder.php
```diff
@@ -42,7 +42,7 @@ class Builder
     /**
      * @var Stage[]
      */
-    private $stages;
+    private $stages = [];
```
venyii: Depending on how long this PR will stay open, could we add this change separately in the next release? Calling `getPipeline` on a new Aggregation Builder will break because of the non-array `$stages` in the `array_map` call.

jmikola: Cross-referencing with https://github.com/doctrine/mongodb/pull/257.

PR #247, opened 2016-01-20
Update documentation link

hugovk:
The "Documentation" link in the READMEs for both these:

 * https://github.com/doctrine/mongodb
 * https://github.com/doctrine/mongodb-odm

Point to:

 * http://docs.doctrine-project.org/projects/doctrine-mongodb-odm/en/latest/index.html

The former should probably link somewhere else to avoid confusion between the two projects. 

Is http://www.doctrine-project.org/projects/mongodb.html the correct one for this project?

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -18,6 +18,6 @@ To run the tests:
 ## More resources:
 
 * [Website](http://www.doctrine-project.org)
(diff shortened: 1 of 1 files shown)


alcaeus:
There is no standalone documentation for this project (apart for the API documentation), which is why the documentation link for the project is a 404.

PR #251, opened 2016-02-29
Corrected fluent interface docblocks

Bilge:
`self` means the base class will always be returned. This is not only false, but breaks code completion in some editors when attempting to access derived class methods. For more information see [phpDocumentor types documentation](https://www.phpdoc.org/docs/latest/guides/types.html#keywords).

### DIFF
--- a/lib/Doctrine/MongoDB/Query/Expr.php
+++ b/lib/Doctrine/MongoDB/Query/Expr.php
@@ -62,7 +62,7 @@ class Expr
      * @see Builder::addAnd()
      * @see http://docs.mongodb.org/manual/reference/operator/and/
      * @param array|Expr $expression
-     * @return self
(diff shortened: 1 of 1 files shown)


malarzm:
Keeping it open for a moment so we remember to change other places, thanks @Bilge for getting started!

alcaeus:
Thanks for getting us started on this, I replaced the remaining occurrences of `@return self` with `@return $this`.

Bilge:
@alcaeus Oh, really? I was just in the middle of doing the same. Did you do it for ODM as well or shall I?

alcaeus:
Apparently, I missed the announcement that you were doing the remaining changes as well - I apologize for that. Go ahead and finish up the ODM pull request. :+1:

Issue #252, opened 2016-03-03
The implementation of the Cursor's current() method always returns the current->file, breaking queryBuilder api


0xPr0xy:
Hello,

I was trying to build a query to select all upload ids when I encountered strange behaviour.
The implementation of the Cursor's current() method always returns the current->file if current is an instance of MongoGridFSFile, then it makes a new GridFSFile of with the data in current.

This transformation does not respect the queryBuilder methods in the example below, and therefore I'm now unable to make a queryBuilder query that selects just the ids of all upload objects.

I'm wondering why this choice was made and if there is a solution to do what I want with the queryBuilder without changing the current() implementation or reverting to raw queries.

```
$queryBuilder = $this->createQueryBuilder();

return $queryBuilder
      ->hydrate(false)
      ->select('_id')
      ->getQuery()
      ->execute();
```

jmikola:
> I'm now unable to make a queryBuilder query that selects just the ids of all upload objects.

How does the problem manifest itself? Is the resulting GridFSFile incomplete because the original query projected only the `_id` field?

> I'm wondering why this choice was made and if there is a solution to do what I want with the queryBuilder without changing the current() implementation or reverting to raw queries.

I don't recall why this design decision was made, but I expect it was done in the interest of user convenience. I could see it conflict if the user already had a "file" field in their `fs.files` document; however, removing this functionality in a minor version would likely cause more headaches as a BC break.

Having looked at #253, I think a compromise would be to respect the query's projection before deciding to inject a GridFSFile object to the `file` property. If the `file` field is excluded either directly or indirectly (i.e. by only selecting other fields), we can skip the behavior. Would that work for you?

alcaeus:
@0xPr0xy could you explain what issues arise from this? I created the following test:
```
    public function testGridFSProjection()
    {
        $gridFS = $this->conn->selectDatabase(self::$dbName)->getGridFS();

        $document = array('foo' => 'bar');
        $gridFS->storeFile(__FILE__, $document);

        $cursor = $gridFS->createQueryBuilder()
            ->select('_id')
            ->getQuery()
            ->execute();

        $item = $cursor->getSingleResult();
    }
```

When running the query with a projection, I get the following item:
```
array(2) {
  '_id' =>
  class MongoId#42 (1) {
    public $$id =>
    string(24) "56eb9c19e78a69d2c40041aa"
  }
  'file' =>
  class Doctrine\MongoDB\GridFSFile#41 (4) {
[...]

PR #253, opened 2016-03-03
current() implementation should not add file by default

0xPr0xy:
Hello,

This is an alternative implementation of the cursor's current() method.
This solves my problem mentioned in issue. #252

### DIFF
--- a/lib/Doctrine/MongoDB/Cursor.php
+++ b/lib/Doctrine/MongoDB/Cursor.php
@@ -151,16 +151,14 @@ class Cursor implements CursorInterface
      * @see http://php.net/manual/en/mongocursor.current.php
      * @return array|null
      */
-    public function current()
-    {
-        $current = $this->mongoCursor->current();
-        if ($current instanceof \MongoGridFSFile) {
-            $document = $current->file;
-            $document['file'] = new GridFSFile($current);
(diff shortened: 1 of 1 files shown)


lib/Doctrine/MongoDB/Cursor.php
```diff
@@ -151,16 +151,14 @@ public function count($foundOnly = false)
      * @see http://php.net/manual/en/mongocursor.current.php
      * @return array|null
      */
-    public function current()
-    {
-        $current = $this->mongoCursor->current();
-        if ($current instanceof \MongoGridFSFile) {
-            $document = $current->file;
-            $document['file'] = new GridFSFile($current);
```
jmikola: See https://github.com/doctrine/mongodb/issues/252#issuecomment-198002727. I think it may be wiser to conditionally skip this injection based on the query's projection.

alcaeus: I agree - that would work best without being a BC break.

alcaeus:
Closing this PR as tests are unstable and the PR needs to target the 1.2.x branch (I'm counting this as a bug which should be fixed in the 1.2.x line as well. I'll get started on the fix with the aforementioned conditional.

PR #255, opened 2016-03-17
Add expr method to aggregation expression object

alcaeus:
This adds the `expr()` method to the expression object the same way it is already available in the builder. This is helpful when needing to create a fresh expression object (to be nested inside) without having access to the actual aggregation builder.

This may be confusing due to the similarity to the `expression()` method that already exists (which can be used to apply an expression previously created with `expr()`) but I can't come up with any more meaningful name.


jmikola:
LGTM. The similar name is unfortunate but I think it's worth keeping this consistent with the Query API.

PR #256, opened 2016-03-18
Allow using operators in group stages

### LINKED ISSUE #254: Aggregation builder does not support arbitrary operators for ID in group stages
Let's take the following aggregation pipeline:
```
[
    { $group: {
        _id: { $dayOfMonth: '$date' },
        signups: { $sum: 1 }
    }}
]
```

In the aggregation builder this stage can't be built since the `Match` class only offers accumulators (`sum`, `avg`, etc.). This was done to make it clear to the user that using any available aggregation operator is not possible. Thus, a workaround using a `$project` stage has to be built:

```
->project()
    ->field('day')
    ->dayOfMonth('$date')
->group()
    ->field('_id')
    ->expression('$day')
    ->field('signups')
    ->sum(1)
```

It would be better if I were able to do this:
```
->group()
    ->field('_id')
    ->dayOfMonth('$date')
    ->field('signups')
    ->sum(1)
```

alcaeus:
Fixes #254.

This change allows users to use operators in group stages for the ID field:
```
$builder
    ->group()
        ->field('_id')
        ->year('$dateField')
        ->field('count')
        ->sum(1);
```

Note: this will also allow building the following pipeline which is invalid and will error out on MongoDB when executed:
```
$builder
    ->group()
        ->field('_id')
        ->year('$dateField')
        ->field('month')
        ->month('$dateField');
```

I've decided against checking for this as this would require overwriting all operators declared in the `Operator` class to check whether the current field is an ID field and error out if it's not. People familiar with aggregation queries should be able to make this distinction on their own.

### DIFF
--- a/lib/Doctrine/MongoDB/Aggregation/Stage/Group.php
+++ b/lib/Doctrine/MongoDB/Aggregation/Stage/Group.php
@@ -29,23 +29,8 @@ use Doctrine\MongoDB\Aggregation\Stage;
  * @author alcaeus <alcaeus@alcaeus.org>
  * @since 1.2
  */
-class Group extends Stage
+class Group extends Operator
 {
-    /**
-     * @var Expr
-     */
-    protected $expr;
-
-    /**
-     * {@inheritdoc}
-     */
-    public function __construct(Builder $builder)
-    {
-        $this->expr = new Expr();
-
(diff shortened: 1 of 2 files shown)


jmikola:
LGTM.

PR #257, opened 2016-03-24
Initialize the aggregation builder stages property with an empty array

venyii:
This will fix a PHP warning when calling `getPipeline()` on an Aggregation
builder without any stages since array_map expects an array as second argument, but the `$stages`
property is still `null`. The new behaviour is consistent with the Query builder, which also
returns an empty array.


alcaeus:
Thanks!

PR #258, opened 2016-03-26
Add PHP 7 to the build matrix using polyfill

alcaeus:
While composer.json has allowed installing on PHP 7 for a while (given that the `ext-mongo` requirement is fulfilled), there were no tests for that platform. This PR adds alcaeus/mongo-php-adapter as polyfill to the PHP 7 version of the build matrix.

Note: due to a [bug with composer](https://github.com/composer/composer/issues/5030) the build step installing the polyfill is running with the `ignore-platform-reqs` flag. Once this bug is fixed the flag should be removed.

Note 2: there is one test that would be failing and is skipped: mainly, it checks whether connecting/disconnecting updates the `connected` flag. Due to the new MongoDB driver not providing explicit connect/disconnect functionality the flag cannot be updated. Since this behavior should not cause too many problems in production I've decided to skip the test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,16 +6,23 @@ php:
   - 5.6
 
 env:
-  - MONGO_VERSION=1.5.8 PREFER_LOWEST=""
-  - MONGO_VERSION=stable PREFER_LOWEST=""
+  - MONGO_VERSION=1.5.8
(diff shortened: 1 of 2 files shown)


malarzm:
:+1:

PR #273, opened 2016-08-31
Removed unused use statements

erheme318:
`grep` result revealed those unused uses while searching on this lib.
So I've corrected it using http://cs.sensiolabs.org/ tool.

Even though it isn't really important, less is better I think. :)


alcaeus:
Thanks!

PR #274, opened 2016-09-06
[feature] Loggable GridFS collection

erheme318:
Problem:
Since GridFS isn't loggable type, no queries are logged. Therefore, it
limits one's capability to optimize queries as investigating logs.

In this PR:
I've introduced `LoggableGridFS` type and added `doGetGridFS` method in
the `LoggableDatabase` in order to initialize loggableGridFS object.

I'm wondering why it wasn't made before in the first place.
If there are some reason, please let me know.


malarzm:
@erheme318 thanks for the PR! We'll take a closer look soon :)

lib/Doctrine/MongoDB/LoggableGridFS.php
```diff
@@ -0,0 +1,74 @@
...
+    {
+        $log['db'] = $this->database->getName();
+        $log['collection'] = $this->getName();
+        call_user_func_array($this->loggerCallable, [$log]);
+    }
+
+    /**
+     * @see Collection::find()
+     */
+    public function find(array $query = [], array $fields = [])
```
erheme318: Other `write` methods can be added if you think it's necessary.

alcaeus: Yes, this should probably log the other methods as well. Keep in mind that the base `GridFS` class extends `Collection`, so this would have to implement the same logging as the collection class itself. What do you think of introducing a `LoggableCollectionTrait` which is used in `LoggableCollection` and `LoggableGridFS` and provides logging functionality for the methods contained in `Collection`?

alcaeus:
Sorry for the delay, I'm not very good at keeping up with pull requests at the moment. To answer your question, yes this was apparently forgotten. I'll gladly merge this once we have all methods covered.

erheme318:
Good to hear that. I'll implement remaining methods then.

lib/Doctrine/MongoDB/Traits/LoggableCollectionTrait.php
```diff
@@ -0,0 +1,362 @@
...
+ * OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ *
+ * This software consists of voluntary contributions made by many individuals
+ * and is licensed under the MIT license. For more information, see
+ * <http://www.doctrine-project.org>.
+ */
+
+namespace Doctrine\MongoDB\Traits;
+
+trait LoggableCollectionTrait
```
alcaeus: I'd suggest adding an `@internal` annotation to signal that this trait is not part of the API.

tests/Doctrine/MongoDB/Tests/LoggableGridFSTest.php
```diff
@@ -0,0 +1,87 @@
...
+        $this->assertEquals($expected, $logResult);
+    }
+
+    public function testStoreFileLog()
+    {
+        $logResult = false;
+
+        $loggableGridFS = $this->getLoggableGridFS($this->getLoggerCallable($logResult));
+
+        $mongoGridFSFile = $this->getMockBuilder('MongoGridFSFile')
```
alcaeus: Since we're targeting PHP 5.6+ with this library, you can use `::class` constants instead of specifying class names in strings.

lib/Doctrine/MongoDB/Traits/LoggableCollectionTrait.php
erheme318: Updated. Thanks

alcaeus:
Thank you very much! 🍻

erheme318:
Thank you too!

erheme318:
Do you have any idea of the release date of `1.4` ?

alcaeus:
Not yet. I'm trying to get #246 into 1.4 but won't be able to get back to it until in a couple of weeks.

erheme318:
Ok. I see

PR #279, opened 2016-11-05
Allow passing multiple arguments to add* methods

### LINKED ISSUE #262: Inconsistency between expression classes
Hi,

When using the mongodb doctrine query builder, I found it hard to transition from the regular doctrine query builder that I'm used to working with in the past.

In Doctrine\MongoDB\Query\Expr , the addAnd method accepts either an Expr instance, or an array.
The andX method in Doctrine\ORM\Query\Expr also accepts an Expr instance or an array. (Well it just does a construct of Expr\Andx with func_get_args) But here, an array of Expr instances is also accepted. 

In the similar class for MongoDB, it accepts an array, which is another format for the expression. Usually Expr::getQuery, which returns an array, is used as argument.

Both methods seem so similar to me, yet they behave differently. Maybe we can do some typechecking on the arguments and throw appropriate exceptions? Also, why wouldn't we call both methods the same if they're supposed to do the same thing? andX vs addAnd i mean..

Here are both methods side by side so you don't have to look them up.

Doctrine\MongoDB\Query\Expr

```
   /**
    * Add an $and clause to the current query.
    * @param array|Expr $expression
    * @return $this
    */
    public function addAnd($expression)
    {
       $this->query['$and'][] = $expression instanceof Expr ? $expression->getQuery() : $expression;
        return $this;
    }
```

Doctrine\ORM\Query\Expr

```
   /**
    * Creates a conjunction of the given boolean expressions.
    *
    * @param \Doctrine\ORM\Query\Expr\Comparison |
    *        \Doctrine\ORM\Query\Expr\Func |
    *        \Doctrine\ORM\Query\Expr\Orx
    *        $x Optional clause. Defaults to null, but requires at least one defined when converting to string.
    *
    * @return Expr\Andx
    */
   public function andX($x = null)
   {
        return new Expr\Andx(func_get_args());
    }
```

alcaeus:
Fixes #262.

This allows passing multiple expressions to the `addAnd`, `addNor` and `addOr` methods in the query builder as well as the `addAnd` and `addOr` methods in the aggregation builder.

### DIFF
--- a/lib/Doctrine/MongoDB/Query/Expr.php
+++ b/lib/Doctrine/MongoDB/Query/Expr.php
@@ -57,16 +57,29 @@ class Expr
     protected $currentField;
 
     /**
-     * Add an $and clause to the current query.
+     * Add one or more $and clauses to the current query.
      *
      * @see Builder::addAnd()
      * @see http://docs.mongodb.org/manual/reference/operator/and/
      * @param array|Expr $expression
      * @return $this
      */
-    public function addAnd($expression)
+    public function addAnd($expression /*, $expression2, ... */)
     {
-        $this->query['$and'][] = $expression instanceof Expr ? $expression->getQuery() : $expression;
+        if (! isset($this->query['$and'])) {
+            $this->query['$and'] = [];
+        }
+
+        $this->query['$and'] = array_merge(
+            $this->query['$and'],
+            array_map(
+                function ($expression) {
+                    return $expression instanceof Expr ? $expression->getQuery() : $expression;
+                },
+                func_get_args()
+            )
+        );
+
         return $this;
     }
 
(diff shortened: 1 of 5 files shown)


lib/Doctrine/MongoDB/Query/Expr.php
```diff
@@ -57,16 +57,24 @@ class Expr
...
+    public function addAnd($expression /*, $expression2, ... */)
     {
-        $this->query['$and'][] = $expression instanceof Expr ? $expression->getQuery() : $expression;
+        if (! isset($this->query['$and'])) {
+            $this->query['$and'] = [];
+        }
+
+        $this->query['$and'] = array_merge(
+            $this->query['$and'],
+            array_map(function ($expression) { return $expression instanceof Expr ? $expression->getQuery() : $expression; }, func_get_args())
```
malarzm: Minor thing but the line is kinda long :)

PR #291, opened 2017-06-14
Allow PHPUnit 6

alcaeus:
This updates the composer constraints to allow PHPUnit 6 on newer systems. There are two risky tests remaining, I'll have to figure out how to deal with them, I guess them not throwing an exception is enough for now.


malarzm:
LGTM, but what's the point?

alcaeus:
> what's the point?

Mainly requiring to 5.7 so I don't have to think which assertion methods to use. At that point I decided to just go ahead and allow 6.*.

malarzm:
👍

PR #292, opened 2017-06-14
Add support for aggregation pipeline stages added in MongoDB 3.4

alcaeus:
This pull requests adds support for the new aggregation pipeline stages added in MongoDB 3.4:
- `$bucket`
- `$bucketAuto`
- `$sortByCount`
- `$replaceRoot`
- `$facet`
- `$graphLookup`
- `$count`

Note: build will fail on PHP 5.6 until I've rebased this on top of #291, but I figured it's worth getting feedback in first.

### DIFF
--- a/lib/Doctrine/MongoDB/Aggregation/Expr.php
+++ b/lib/Doctrine/MongoDB/Aggregation/Expr.php
@@ -90,7 +90,7 @@ class Expr
             $this->expr['$and'] = [];
         }
 
-        $this->expr['$and'] = array_merge($this->expr['$and'], array_map([$this, 'ensureArray'], func_get_args()));
+        $this->expr['$and'] = array_merge($this->expr['$and'], array_map(['static', 'convertExpression'], func_get_args()));
 
         return $this;
     }
@@ -108,7 +108,7 @@ class Expr
             $this->expr['$or'] = [];
         }
 
-        $this->expr['$or'] = array_merge($this->expr['$or'], array_map([$this, 'ensureArray'], func_get_args()));
+        $this->expr['$or'] = array_merge($this->expr['$or'], array_map(['static', 'convertExpression'], func_get_args()));
 
         return $this;
     }
@@ -281,25 +281,38 @@ class Expr
     }
 
     /**
+     * Converts an expression object into an array, recursing into nested items
+     *
+     * For expression objects, it calls getExpression on the expression object.
+     * For arrays, it recursively calls itself for each array item. Other values
+     * are returned directly
+     *
+     * @param mixed|self $expression
+     * @return string|array
+     * @internal
+     */
+    public static function convertExpression($expression)
+    {
+        if (is_array($expression)) {
+            return array_map(['static', 'convertExpression'], $expression);
+        } elseif ($expression instanceof self) {
+            return $expression->getExpression();
+        }
+
+        return $expression;
+    }
+
+    /**
      * Ensures an array or operator expression is converted to an array.
+     *
+     * @deprecated Deprecated in favor of convertExpression
      *
      * @param mixed|self $expression
      * @return mixed
      */
     protected function ensureArray($expression)
     {
-        if (is_array($expression)) {
-            $array = [];
-            foreach ($expression as $index => $value) {
-                $array[$index] = $this->ensureArray($value);
-            }
-
-            return $array;
-        } elseif ($expression instanceof self) {
-            return $expression->getExpression();
-        }
-
-        return $expression;
+        return static::convertExpression($expression);
     }
 
     /**
@@ -430,7 +443,7 @@ class Expr
     public function expression($value)
     {
         $this->requiresCurrentField(__METHOD__);
-        $this->expr[$this->currentField] = $this->ensureArray($value);
+        $this->expr[$this->currentField] = static::convertExpression($value);
 
         return $this;
     }
@@ -881,9 +894,9 @@ class Expr
     private function operator($operator, $expression)
     {
         if ($this->currentField) {
-            $this->expr[$this->currentField][$operator] = $this->ensureArray($expression);
+            $this->expr[$this->currentField][$operator] = static::convertExpression($expression);
         } else {
-            $this->expr[$operator] = $this->ensureArray($expression);
+            $this->expr[$operator] = static::convertExpression($expression);
         }
 
         return $this;
--- a/lib/Doctrine/MongoDB/Aggregation/Expr.php
+++ b/lib/Doctrine/MongoDB/Aggregation/Expr.php
@@ -285,7 +285,7 @@ class Expr
      *
      * For expression objects, it calls getExpression on the expression object.
      * For arrays, it recursively calls itself for each array item. Other values
-     * are returned directly
+     * are returned directly.
      *
      * @param mixed|self $expression
      * @return string|array
--- a/lib/Doctrine/MongoDB/Aggregation/Stage/AbstractBucket.php
+++ b/lib/Doctrine/MongoDB/Aggregation/Stage/AbstractBucket.php
@@ -1,0 +1,60 @@
+<?php
+
+namespace Doctrine\MongoDB\Aggregation\Stage;
+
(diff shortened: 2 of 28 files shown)


lib/Doctrine/MongoDB/Aggregation/Expr.php
```diff
@@ -281,17 +281,22 @@ public function cond($if, $then, $else)
...
+     * @internal
      */
-    protected function ensureArray($expression)
+    public static function convertExpression($expression)
     {
         if (is_array($expression)) {
             $array = [];
             foreach ($expression as $index => $value) {
-                $array[$index] = $this->ensureArray($value);
+                $array[$index] = static::convertExpression($value);
```
jmikola: Is `$expression` always a compact array in this case, or might it be associative? I'm just curious if `$index` is relevant here or if `$array[]` would be functionally equivalent.

lib/Doctrine/MongoDB/Aggregation/Stage/AbstractBucket.php
```diff
@@ -0,0 +1,59 @@
...
+ * @author alcaeus <alcaeus@alcaeus.org>
+ * @since 1.5
+ */
+abstract class AbstractBucket extends Stage
+{
+    /**
+     * @var Bucket\BucketOutput|null
+     */
+    protected $output;
+    /**
```
jmikola: Do you need a newline above this?

lib/Doctrine/MongoDB/Aggregation/Stage/Bucket.php
```diff
@@ -0,0 +1,89 @@
...
+     * @var mixed
+     */
+    private $default;
+
+    /**
+     * @param array ...$boundaries
+     *
+     * @return $this
+     */
+    public function boundaries(...$boundaries)
```
jmikola: Should this validate the array structure (per [docs](https://docs.mongodb.com/manual/reference/operator/aggregation/bucket/)), or do you generally rely on the server to report an error?

lib/Doctrine/MongoDB/Aggregation/Stage/ReplaceRoot.php
```diff
@@ -0,0 +1,58 @@
...
+        $this->expression = $expression;
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public function getExpression()
+    {
+        return [
+            '$replaceRoot' => $this->expression ? Expr::convertExpression($this->expression) : $this->expr->getExpression()
```
jmikola: Would `$this->expression !== null` be safer here, on the off chance the string is falsey?

tests/Doctrine/MongoDB/Tests/Aggregation/Stage/BucketAutoTest.php
```diff
@@ -0,0 +1,62 @@
...
+namespace Doctrine\MongoDB\Tests\Aggregation\Stage;
+
+use Doctrine\MongoDB\Aggregation\Stage\BucketAuto;
+use Doctrine\MongoDB\Tests\Aggregation\AggregationTestCase;
+
+class BucketAutoTest extends \PHPUnit_Framework_TestCase
+{
+    use AggregationTestCase;
+
+    public function testBucketStage()
```
jmikola: Should all three of these test methods start with "testBucketAuto"?

lib/Doctrine/MongoDB/Aggregation/Expr.php
alcaeus: An expression could be this:
```
[
    'avgField' => ['$avg' => '$field'],
    'sumField' => ['$sum' => '$field'],
]
```

That said, I did want to change it to an `array_map`, so thanks for reminding me!

lib/Doctrine/MongoDB/Aggregation/Stage/Bucket.php
alcaeus: I've tried to leave out as much validation logic as possible and leave it to the server/driver to report any errors.

lib/Doctrine/MongoDB/Aggregation/Stage/ReplaceRoot.php
alcaeus: You're right. I'll change it.

tests/Doctrine/MongoDB/Tests/Aggregation/Stage/BucketAutoTest.php
alcaeus: Copy/paste busted ;)

lib/Doctrine/MongoDB/Aggregation/Stage/Facet.php
```diff
@@ -0,0 +1,86 @@
...
+        ];
+    }
+
+    /**
+     * @param string $fieldName
+     * @return $this
+     */
+    public function field($fieldName)
+    {
+        $this->fieldName = $fieldName;
```
malarzm: Why the difference in name?

lib/Doctrine/MongoDB/Aggregation/Stage/Project.php
```diff
@@ -61,11 +61,15 @@ public function avg($expression1/* , $expression2, ... */)
 
     /**
      * Shorthand method to exclude the _id field.
+     *
+     * @deprecated Deprecated in 1.5, please use @see excludeFields.
```
malarzm: Is this correct usage of `@see`?

lib/Doctrine/MongoDB/Aggregation/Stage/SortByCount.php
```diff
@@ -0,0 +1,60 @@
...
+     */
+    public function __construct(Builder $builder, $fieldName)
+    {
+        parent::__construct($builder);
+
+        $this->fieldName = (string) $fieldName;
+    }
+
+    /**
+     * Assembles the aggregation stage
```
malarzm: Missing "." :)

lib/Doctrine/MongoDB/Aggregation/Expr.php
```diff
@@ -281,17 +281,22 @@ public function cond($if, $then, $else)
...
+     * For arrays, it recursively calls itself for each array item. Other values
+     * are returned directly
      *
      * @param mixed|self $expression
-     * @return mixed
+     * @return string|array
+     * @internal
      */
-    protected function ensureArray($expression)
+    public static function convertExpression($expression)
```
alcaeus: Yes, that's the main reason for the refactoring: it's used in a couple of the new stages, so to avoid duplicating the behavior I decided to make it a public method instead. It's marked internal since there's no need to use it in applications.

lib/Doctrine/MongoDB/Aggregation/Stage/Bucket.php
```diff
@@ -0,0 +1,89 @@
...
+    {
+        $this->boundaries = $boundaries;
+        return $this;
+    }
+
+    /**
+     * @param mixed $default
```
alcaeus: Yes, as this can be any value. From the documentation:
> A literal that specifies the _id of an additional bucket that contains all documents whose groupBy expression result does not fall into a bucket specified by boundaries.

As such I thought it best to make it `mixed`.

lib/Doctrine/MongoDB/Aggregation/Stage/Project.php
alcaeus: I'm not sure - I was hoping it is :)

alcaeus:
@malarzm took care of your comments and added descriptions to most methods in new stages. 👍

lib/Doctrine/MongoDB/Aggregation/Stage/Project.php
malarzm: I think you should use `@link` instead - http://manual.phpdoc.org/HTMLSmartyConverter/HandS/phpDocumentor/tutorial_tags.see.pkg.html

alcaeus: Done 👍

PR #293, opened 2017-06-26
Add aggregation operators added in MongoDB 3.4

alcaeus:
This adds almost all new aggregation operators introduced in MongoDB 3.4 to the aggregation builder. Still missing:
* [x] [`$collStats`](https://docs.mongodb.com/manual/reference/operator/aggregation/collStats/): I'm not sure what a smart API would be, but most likely I'd make the signature something like `collStats(bool $latencyStats, bool $latencyHistograms, bool $histograms)` and construct the expression from the three bools given.
* [x] [`$switch`](https://docs.mongodb.com/manual/reference/operator/aggregation/switch/): this would require using different names for `case`, `then` and `default` as those are reserved words. Only when using PHP 7 can we use these words as method names. One option would be to add other methods and mark them as deprecated from the outset.


lib/Doctrine/MongoDB/Aggregation/Expr.php
```diff
@@ -584,6 +584,103 @@ public function ifNull($expression, $replacementExpression)
...
+     */
+    public function indexOfArray($arrayExpression, $searchExpression, $start = null, $end = null)
+    {
+        $args = [$arrayExpression, $searchExpression];
+        if ($start !== null) {
+            $args[] = $start;
+        }
+        if ($end !== null) {
+            $args[] = $end;
+        }
```
jmikola: Should it be a logical error if `$start` is null but `$end` is non-null? Otherwise, `$end` will be appended to `$args` in the position where `$start` is expected.

lib/Doctrine/MongoDB/Aggregation/Expr.php
```diff
@@ -584,6 +584,103 @@ public function ifNull($expression, $replacementExpression)
...
+     */
+    public function indexOfBytes($stringExpression, $substringExpression, $start = null, $end = null)
+    {
+        $args = [$stringExpression, $substringExpression];
+        if ($start !== null) {
+            $args[] = $start;
+        }
+        if ($end !== null) {
+            $args[] = $end;
+        }
```
jmikola: Similar issue here if `$start` is null and `$end` is non-null. Looks like there are other cases below, but I'll refrain from spamming.

tests/Doctrine/MongoDB/Tests/Aggregation/Stage/OperatorTest.php
```diff
@@ -3,103 +3,52 @@
...
 
 use Doctrine\MongoDB\Aggregation\Expr;
+use Doctrine\MongoDB\Tests\Aggregation\AggregationOperatorsProviderTrait;
 use Doctrine\MongoDB\Tests\Aggregation\AggregationTestCase;
 use Doctrine\MongoDB\Tests\TestCase;
 
 class OperatorTest extends TestCase
 {
-    use AggregationTestCase;
+    use AggregationTestCase, AggregationOperatorsProviderTrait;
```
jmikola: Clever use of traits for test provisioning :)

jmikola:
> `$switch`: this would require using different names for case, then and default as those are reserved words. Only when using PHP 7 can we use these words as method names. One option would be to add other methods and mark them as deprecated from the outset.

I'm curious if you could get around this by documenting those methods and having PHP < 7 handle them via `__call()`. Is the reserved word validation only when defining them as methods in PHP 5.x source, or also when invoking the names?

lib/Doctrine/MongoDB/Aggregation/Expr.php
alcaeus: Yes, that makes sense. I'll add that to all methods. Didn't think about our method signature allowing that use case, great catch! 👍

alcaeus:
> I'm curious if you could get around this by documenting those methods and having PHP < 7 handle them via `__call()`. Is the reserved word validation only when defining them as methods in PHP 5.x source, or also when invoking the names?

That is a very clever idea, and it seems to work: https://3v4l.org/GdgJs. I'm wondering whether it would make sense to extract the logic to a trait and conditionally implement this using `__call` for PHP 5.6 and just straight-up declare the methods for PHP 7.

alcaeus:
@jmikola @malarzm Added the `$collStats` pipeline stage that I missed in the other pull request (I thought it was just an operator, don't know why) and also included a minor refactoring to use `$this` as return type in the docblock. This makes code completion more useful in ODM where we extend the `Expr` class and add our own methods.

lib/Doctrine/MongoDB/Aggregation/Builder.php
```diff
@@ -115,6 +115,22 @@ public function bucketAuto()
     }
 
     /**
+     * Returns statistics regarding a collection or view.
+     *
+     * $collStats must be the first stage in an aggregation pipeline, or else
```
malarzm: Would it be feasible to throw an exception in builder already?

lib/Doctrine/MongoDB/Aggregation/Expr.php
```diff
@@ -194,6 +219,26 @@ public function avg($expression)
...
+     *
+     * Requires {@link switch()} to be called first. The argument can be any
+     * valid expression that resolves to a boolean. If the result is not a
+     * boolean, it is coerced to a boolean value.
+     *
+     * @param mixed|self $expression
+     *
+     * @return $this
+     */
+    protected function caseInternal($expression)
```
malarzm: Should this method be private?

lib/Doctrine/MongoDB/Aggregation/Expr.php
```diff
@@ -950,6 +1190,54 @@ private function requiresCurrentField($method = null)
...
+     * @param string $method
+     *
+     * @throws \BadMethodCallException if there is no current switch operator
+     */
+    private function requiresSwitchStatement($method = null)
+    {
+        $message = ($method ?: 'This method') . ' requires a valid switch statement (call switch() first).';
+
+        if ($this->currentField) {
+            if (!isset($this->expr[$this->currentField]['$switch'])) {
```
malarzm: Are we no longer putting space after `!` or is this `doctrine/mongodb` only?

tests/Doctrine/MongoDB/Tests/Aggregation/Stage/OperatorTest.php
malarzm: :+1:

lib/Doctrine/MongoDB/Aggregation/Expr.php
alcaeus: No - since I consider it part of the public interface I thought it best to allow overriding it just like all other stage methods as well.

lib/Doctrine/MongoDB/Aggregation/Expr.php
alcaeus: This might be me leaking coding standards from other projects into here - I'd have to check. If you ask me I'd do away with the space 😛

lib/Doctrine/MongoDB/Aggregation/Expr.php
malarzm: As per IRC I'm just leaving a reminder to add a deprecation notice already

lib/Doctrine/MongoDB/Aggregation/Builder.php
alcaeus: I'm trying to leave out as much of this server specific logic as possible. That's why I haven't added it for `$geoNear` which has the same restriction. (see https://github.com/doctrine/mongodb/pull/292#discussion_r122291605)

alcaeus:
@malarzm will clean up coding style in a separate pull request, there are a few more instances where it's wrong.

PR #297, opened 2017-07-06
Enforce primary read preference in query builder for write commands

jmikola:
Related to:

 * https://github.com/alcaeus/mongo-php-adapter/issues/181
 * https://github.com/alcaeus/mongo-php-adapter/issues/180
 * https://github.com/mongodb/mongo-php-library/issues/384

This mainly fixes an incompatibility with `mongo-php-adapter`, since the legacy `mongo` extension already forces a primary read preference for most commands (as discussed in https://github.com/alcaeus/mongo-php-adapter/issues/181).

Of note, the legacy driver never forces primary read preference for the `aggregate` command, as it does not inspect the pipeline for an `$out` stage. The code path for this is also more complicated, as Doctrine MongoDB delegates to `MongoDB::command()` or `MongoCollection::aggregateCursor()` depending on whether an output cursor is requested. Doctrine MongoDB's aggregation builder also does not take a read preference option, so users will be forced to explicitly call `Collection::setReadPreference('primary')` before constructing an aggregation pipeline that includes `$out` if they are operating within the context of a read preference that could select a secondary server.

That said, I'm inclined to say that it's not worth the trouble to address the aggregation case given that users can work around this if needed.

### DIFF
--- a/lib/Doctrine/MongoDB/Query/Query.php
+++ b/lib/Doctrine/MongoDB/Query/Query.php
@@ -174,20 +174,28 @@ class Query implements IteratorAggregate
                 $queryOptions = $this->getQueryOptions('new', 'select', 'sort', 'upsert');
                 $queryOptions = $this->renameQueryOptions($queryOptions, ['select' => 'fields']);
 
-                return $this->collection->findAndUpdate(
-                    $this->query['query'],
-                    $this->query['newObj'],
-                    array_merge($options, $queryOptions)
-                );
+                $closure = function() use ($options, $queryOptions) {
+                    return $this->collection->findAndUpdate(
+                        $this->query['query'],
+                        $this->query['newObj'],
+                        array_merge($options, $queryOptions)
+                    );
+                };
+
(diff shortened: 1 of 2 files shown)


lib/Doctrine/MongoDB/Query/Query.php
```diff
@@ -174,20 +174,34 @@ public function execute()
...
 
-                return $this->collection->findAndUpdate(
-                    $this->query['query'],
-                    $this->query['newObj'],
-                    array_merge($options, $queryOptions)
-                );
+                $collection = $this->collection;
+                $query = $this->query;
+
+                $closure = function() use ($collection, $query, $options, $queryOptions) {
```
alcaeus: FWIW, you can use `$this` in closures (support for binding `$this` automatically was added in PHP 5.4).

jmikola: Will do. This was due to the old habit of supporting PHP 5.3, although the library's requirement is well beyond that now :)

PR #298, opened 2017-07-18
Fix BC break when converting expression objects

alcaeus:
This was introduced in #292 by adding a `convertExpression` method and using that in the expression operator methods. Unfortunately, MongoDB ODM relied on the `ensureArray` method being called to do the type conversion, which is no longer the case since all methods call `convertExpression`.

### DIFF
--- a/lib/Doctrine/MongoDB/Aggregation/Stage/AbstractBucket.php
+++ b/lib/Doctrine/MongoDB/Aggregation/Stage/AbstractBucket.php
@@ -45,7 +45,7 @@ abstract class AbstractBucket extends Stage
     {
         $stage = [
             '$bucket' => [
-                'groupBy' => Expr::convertExpression($this->groupBy),
+                'groupBy' => $this->convertExpression($this->groupBy),
             ] + $this->getExtraPipelineFields(),
         ];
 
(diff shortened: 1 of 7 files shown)


lib/Doctrine/MongoDB/Aggregation/Stage/AbstractBucket.php
```diff
@@ -57,6 +57,21 @@ public function getExpression()
...
+     * Converts an expression object into an array, recursing into nested items
+     *
+     * For expression objects, it calls getExpression on the expression object.
+     * For arrays, it recursively calls itself for each array item. Other values
+     * are returned directly.
+     *
+     * @param mixed|self $expression
+     * @return string|array
+     */
+    protected function convertExpression($expression)
```
alcaeus: Introducing this method is necessary to allow mongodb-odm to override it and convert the expression using classMetadata from the document the aggregation builder is using. This would require us to release this bugfix as 1.6.0 since protected methods in non-final classes are part of the public API. However, I couldn't find a good way to solve this issue without introducing another method. What do you think @jmikola @malarzm?

lib/Doctrine/MongoDB/Aggregation/Stage/AbstractBucket.php
```diff
@@ -57,6 +57,21 @@ public function getExpression()
     }
 
     /**
+     * Converts an expression object into an array, recursing into nested items
+     *
+     * For expression objects, it calls getExpression on the expression object.
+     * For arrays, it recursively calls itself for each array item. Other values
+     * are returned directly.
```
jmikola: Does this comment still apply? I wonder if it makes more sense to explain the purpose of this method as an extendable hook, rather than repeating the documentation for the method it proxies.

lib/Doctrine/MongoDB/Aggregation/Stage/AbstractBucket.php
jmikola: I concur with a version bump, unfortunate as it is.

lib/Doctrine/MongoDB/Aggregation/Stage/AbstractBucket.php
alcaeus: Good point, will update.

PR #306, opened 2017-10-04
Fix wrong name for $bucketAuto pipeline stage

alcaeus:
This fixes a rather silly copy/paste mistake on my part. I'll note that this fix technically constitutes a BC break, as we're adding a new abstract method to a class. Given the lack of bug reports and the fact that ODM 1.2 with support for the aggregation framework still hasn't hit yet, i'm willing to push the fix this way. Other option would be to have `getStageName` non-abstract and return `$bucket` in `AbstractBucket` and only override the method in `BucketAuto`.

### DIFF
--- a/lib/Doctrine/MongoDB/Aggregation/Stage/AbstractBucket.php
+++ b/lib/Doctrine/MongoDB/Aggregation/Stage/AbstractBucket.php
@@ -44,13 +44,13 @@ abstract class AbstractBucket extends Stage
     public function getExpression()
     {
         $stage = [
-            '$bucket' => [
+            $this->getStageName() => [
                 'groupBy' => $this->convertExpression($this->groupBy),
             ] + $this->getExtraPipelineFields(),
         ];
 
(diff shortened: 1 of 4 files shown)


lib/Doctrine/MongoDB/Aggregation/Stage/AbstractBucket.php
```diff
@@ -75,4 +75,11 @@ protected function convertExpression($expression)
      * @return array
      */
     abstract protected function getExtraPipelineFields();
+
+    /**
+     * Returns the stage name with the dollar prefix
+     *
+     * @return string
+     */
+    abstract protected function getStageName();
```
jmikola: AbstractBucket is annotated with `@internal` so I wouldn't even consider this a BC break.

alcaeus: Oh, I missed that completely. Even better! 👍

PR #308, opened 2017-10-16
Allow passing bucket boundaries as an array as well

metanav:
Sometimes we have use cases when boundaries are created dynamically or read from config files.


alcaeus:
> Sometimes we have use cases when boundaries are created dynamically or read from config files.

Then use the splat operator when calling `boundaries`:
```
$stage
    ->boundaries(...$dynamicallyGeneratedArrayWithBoundaries)
    ->...
;
```

The interface is fine as is - I don't want multiple ways to call a single method. Closing as invalid.

metanav:
Oh! Didn't realize that splat should work.  Thanks!

PR #312, opened 2017-12-05
Refactoring tests

carusogabriel:
Continuing to refactory tests, like #309, I used:
- `assertInternalType` instead of `is_*` functions;
- `assertCount` instead of `count` function;
- `assertArrayNotHasKey` instead of `array_key_exists` function.


alcaeus:
Thanks @carusogabriel!

PR #320, opened 2018-03-07
Fix some phpstan warnings

vkhramtsov:


alcaeus:
Thanks for your contribution, @vkhramtsov! Unfortunately, this library is in maintenance-only mode, so I won't be merging this. The code has been merged into `doctrine/mongodb-odm`, where we also made similar changes and added phpstan to our CI build: https://github.com/doctrine/mongodb-odm/pull/1714. We're currently running on a low level, so if you want to increase the level and fix some errors over in the ODM repository, we'd really appreciate it!

vkhramtsov:
@alcaeus Would you be so kind to write about maintenance-mode only in readme? Just to prevent creation PR's like this.

alcaeus:
Of course. Just did so in 94844dcaa39d2bad3189bd924f20d6a67c8ca424.

vkhramtsov:
Thank you

PR #333, opened 2019-07-09
Fix wrong syntax for replaceRoot stage

alcaeus:
This PR fixes a wrong syntax for replaceRoot, as reported in https://github.com/doctrine/DoctrineMongoDBBundle/issues/565.

### DIFF
--- a/lib/Doctrine/MongoDB/Aggregation/Stage/ReplaceRoot.php
+++ b/lib/Doctrine/MongoDB/Aggregation/Stage/ReplaceRoot.php
@@ -4,6 +4,7 @@ namespace Doctrine\MongoDB\Aggregation\Stage;
 
 use Doctrine\MongoDB\Aggregation\Builder;
 use Doctrine\MongoDB\Aggregation\Expr;
+use function is_array;
 
 /**
  * Fluent interface for adding a $replaceRoot stage to an aggregation pipeline.
@@ -35,8 +36,12 @@ class ReplaceRoot extends Operator
      */
     public function getExpression()
     {
+        $expression = $this->expression !== null ? $this->convertExpression($this->expression) : $this->expr->getExpression();
+
         return [
-            '$replaceRoot' => $this->expression !== null ? $this->convertExpression($this->expression) : $this->expr->getExpression()
+            '$replaceRoot' => [
+                'newRoot' => is_array($expression) ? (object) $expression : $expression,
+            ],
         ];
     }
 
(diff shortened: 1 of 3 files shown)


lib/Doctrine/MongoDB/Aggregation/Stage/ReplaceRoot.php
```diff
@@ -36,7 +36,9 @@ public function __construct(Builder $builder, $expression = null)
     public function getExpression()
     {
         return [
-            '$replaceRoot' => $this->expression !== null ? $this->convertExpression($this->expression) : $this->expr->getExpression()
+            '$replaceRoot' => [
+                'newRoot' => $this->expression !== null ? $this->convertExpression($this->expression) : $this->expr->getExpression(),
```
jmikola: Looking at [`$replaceRoot`](https://docs.mongodb.com/manual/reference/operator/aggregation/replaceRoot/), `newRoot` should be "any valid expression that resolves to a document."  `ReplaceRoot::convertExpression()` is documented as returning a string or array. The doc block for `Expression::getExpression()` suggests it always returns an array.

Two things to consider here:

 * Will we ever end up with a string value for the expression? If so, should we raise an error preemptively or allow the server to do so?
 * If the expression is empty or has numeric fields that make it indistinguishable from a packed array, we could end up passing a BSON array to the server instead of a document. Recent server versions have gotten stricter about validating BSON types and might report an error in this case. We can easily work around that by casting the array to an `(object)`.

alcaeus: > Will we ever end up with a string value for the expression? If so, should we raise an error preemptively or allow the server to do so?

We might end up with a string value for the expression, e.g. by using `replaceRoot()->expression('$$ROOT')`. However, this expression resolves to a document, working as expected. Since we can't really resolve the expressions given, it's probably best to defer this kind of check to the server.

> If the expression is empty or has numeric fields that make it indistinguishable from a packed array, we could end up passing a BSON array to the server instead of a document. Recent server versions have gotten stricter about validating BSON types and might report an error in this case. We can easily work around that by casting the array to an `(object)`.

This is a good point. We already do so in `$match` and `graphLookup` pipeline stages if no query was given. A cast to `object` would probably solve most issues even though it won't recurse into nested arrays. This is something we might want to change in ODM 2.x somewhere down the line 👍

jmikola: > However, this expression resolves to a document, working as expected. Since we can't really resolve the expressions given, it's probably best to defer this kind of check to the server.

Ah, very good point. I did not consider that variable syntax, where a string in the client could evaluate as a document server-side.

PR #337, opened 2019-08-30
Make requirements compatible with Debian

dvdknaap:


malarzm:
This library can't work with the new driver and will not be updated to make it compatible. If you need to use ODM 1.x please follow steps described in https://www.doctrine-project.org/projects/doctrine-mongodb-odm/en/1.2/reference/introduction.html#using-php-7

PR #342, opened 2020-08-06
Drop doctrine common dependency

### LINKED ISSUE #341: Drop doctrine/common dependency
This library relies on doctrine/common repository.

Even tough this library is deprecated, would it still be possible to drop the doctrine/common dependency in favor of the replacement repositories like doctrine/collections and doctrine/persistence?

It would enable one to upgrade other dependencies relying on doctrine/common >= 3.0.

zluiten:
Close #341

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,7 @@
     "require": {
         "php": "^5.6 || ^7.0",
         "ext-mongo": "^1.6.7",
-        "doctrine/common": "^2.2"
+        "doctrine/event-manager": "^1.0"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


SenseException:
@alcaeus Is `master` going to drop PHP <= 7.0?

alcaeus:
No, that can't happen due to the dependency on ext-mongo which isn't available on PHP 7.

alcaeus:
Thanks @netiul for the attempt to fix this. However, as this library is no longer actively maintained, I'm closing this PR.

